PR #29, opened 2018-07-04
add staging environments for Form Builder

aldavidson:


namespaces/non-production.k8s.integration.dsd.io/formbuilder-services-staging/formbuilder-services-staging-admin-role.yaml
```diff
@@ -0,0 +1,64 @@
+# Role to allow listing of namespaces.
```
alkar: Same as above.

namespaces/non-production.k8s.integration.dsd.io/formbuilder-platform-staging/formbuilder-platform-staging-admin-role.yaml
```diff
@@ -0,0 +1,64 @@
+# Role to allow listing of namespaces.
```
alkar: The `-namespaces-list` and `-basic-user` `ClusterRole` and `ClusterRoleBindings` are not needed here, they are already applied for all users on the cluster level.

PR #33, opened 2018-07-09
Add terraform

jojuolape:


whichNamespace.sh
```diff
@@ -10,5 +10,15 @@ else
     kubectl config use-context $CLUSTER
     echo "Running python3 namespace.py -c $CLUSTER"
     python3 namespace.py -c $CLUSTER
+    for file in namespaces/*; do
```
alkar: This is looping through the clusters, while the loop in L8 does the same. Wouldn't this apply terraform multiple times on each run?

namespaces/cloud-platform-test-1.k8s.integration.dsd.io/platforms-dev/resources/main.tf
```diff
@@ -0,0 +1,17 @@
+provider "aws" {
```
alkar: Should we add the backend config here?
This will be the first example of how to use terraform-managed resources so I think we should manage the state as well.

alkar:
Closing, superseded by #34

PR #38, opened 2018-07-13
Prometheus and alertmanager oauth2 proxy

alkar:
Connects to https://github.com/ministryofjustice/cloud-platform/issues/161


alkar:
@kerin that a good point, I picked this up last night and just went with the flow - this solution a  feature of `ingress-nginx` (https://github.com/kubernetes/ingress-nginx/tree/master/docs/examples/auth/oauth-external-auth) which feels a bit obscure.

Do you have any suggestions for an oidc proxy?

kerin:
Yeah, doing it with nginx-ingress seems weird to me. We could potentially do it with Istio (I think) or Ambassador - https://www.getambassador.io/ - but there’s obviously a lot to think about there. For a straight swap from an oauth or OIDC proxy, this looks good - https://github.com/evry/docker-oidc-proxy - and there’s also a nodeJS one we made on Analytics because we couldn’t find a good existing one at the time: https://github.com/ministryofjustice/analytics-platform-auth-proxy

alkar:
Closing in favour of #40

PR #59, opened 2018-07-31
Fluentd dir and 2 files for the DaemonSet and RBAC

ollieanwyll:
This PR is to confirm that the 2 Fluentd DaemonSet and RBAC files are correct in their functionality and syntax.


namespaces/cloud-platform-test-1.k8s.integration.dsd.io/fluentd/fluentd-ds-rbac.yaml
```diff
@@ -0,0 +1,37 @@
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: fluentd
+  namespace: kube-system
```
alkar: All of these resources are deployed in `kube-system` but the files themselves are under `fluentd`. Can you please adjust?

We'd want `fluentd` to run on its own namespace (perhaps called `logs` or similar).

PR #64, opened 2018-08-06
Abdi creating environment userguide

AbdirizakObsiye2018:
I've added the cloud platform reference app and also a rbac-config file to allow tiller to have admin access into my namespace.


jasonBirchall:
This is a training PR and will be closed following a chat with the engineer.

PR #72, opened 2018-08-08
Add AWS resources with terraform

rowlando:


namespaces/cloud-platform-live-0.k8s.integration.dsd.io/nick-sample-app-dev/resources/ecr.tf
```diff
@@ -0,0 +1,27 @@
...
+}
+
+provider "aws" {
+  region = "eu-west-1"
+}
+
+module "ecr-repo" {
+  source = "github.com/ministryofjustice/cloud-platform-terraform-ecr-credentials?ref=master"
+
+  team_name = "my-team-name"
```
alkar: Can you please adjust `team_name` and `repo_name` to some value that is different from the documentation example?

namespaces/cloud-platform-live-0.k8s.integration.dsd.io/nick-sample-app-dev/resources/ecr.tf
```diff
@@ -0,0 +1,27 @@
...
+  repo_name = "my-app-name"
+}
+
+resource "kubernetes_secret" "ecr-repo" {
+  metadata {
+    name      = "ecr-repo-my-app-name"
+    namespace = "my-environment"
```
alkar: This should match the namespace you want the values to be made accessible to `nick-sample-app-dev` in this case.

PR #73, opened 2018-08-08
add ns resource folder with ns creation files

sablumiah:
WHAT
Created a new folder with mandatory files that are required to create a new ns including new files for limitrange and resourcequota

WHY
limitrange and resourcequota are required so we can have better sizing and node utilisation.
The values are based on live-0 node sizes of c4 2xl. These will need to be adjusted when using on lower spec clusters

### DIFF
--- a/namespace-resources/02-limitrange.yaml
+++ b/namespace-resources/02-limitrange.yaml
@@ -1,0 +1,13 @@
+apiVersion: v1
+kind: LimitRange
+metadata:
+  name: limitrange
+spec:
+  limits:
+  - default:
+      cpu: 2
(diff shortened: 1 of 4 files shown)


namespace-resources/02-limitrange.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
+kind: LimitRange
+metadata:
+  name: limitrange
+spec:
+  limits:
+  - default:
+      cpu: 2
```
alkar: Should we use millicpus throughout for consistency?

PR #105, opened 2018-08-24
Add new default Network Policy yaml file to resources folder

sablumiah:
New default network policy created 
Network policy prohibits all traffic from one namespace to another except for ingress-controllers namespaces to all other namespaces

Added label to ingress controller NS - component=ingress-controllers

networkpolicy.md file created

### DIFF
--- a/docs/networkpolicy.md
+++ b/docs/networkpolicy.md
@@ -1,5 +1,25 @@
 # Network Policy
 
-ABC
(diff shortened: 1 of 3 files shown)


jasonBirchall:
I believe it was agreed to hold off putting this into live until it has been thoroughly tested in test-1 (which it has not). 

This is referenced in this issue: https://github.com/ministryofjustice/cloud-platform/issues/342

PR #109, opened 2018-08-30
Create the Big Book of Acronyms' namespace

SteveMarshall:
I'm investigating migrating the [Big Book of Acronyms](https://github.com/ministryofjustice/bba/) from [Heroku](https://bigbookofacronyms.herokuapp.com) (where it's running out of free compute) to the Cloud Platform.

A few things to note:

- We don't need multiple environments, because it's only really looked after by me, and isn't really business-critical.
- I've intentionally omitted a `RoleBinding` because I'm not sure that anyone other than me is really in a position to look after this, and I get access by virtue of being a member of the Cloud Platform team. We might want to reconsider that, though, and add an explicit owner.
- I've tried to adhere to [our standards for documenting owners of infrastructure](https://ministryofjustice.github.io/technical-guidance/standards/documenting-infrastructure-owners/). I've set the `owner` and `infrastructure-support` to me; should they perhaps point to the Cloud Platform team?

And, one thing to discuss is the use of `cloud-platform.justice.gov.uk` as our cluster-wide label/annotation prefix. Does anyone already use something, or should we standardise on that?

### DIFF
--- a/namespaces/cloud-platform-live-0.k8s.integration.dsd.io/big-book-of-acronyms/namespace.yaml
+++ b/namespaces/cloud-platform-live-0.k8s.integration.dsd.io/big-book-of-acronyms/namespace.yaml
@@ -4,11 +4,11 @@ metadata:
   name: big-book-of-acronyms
   labels:
     name: big-book-of-acronyms
-    business-unit: platforms
-    is-production: true
-    environment-name: production
+    cloud-platform.justice.gov.uk/business-unit: platforms
+    cloud-platform.justice.gov.uk/is-production: true
+    cloud-platform.justice.gov.uk/environment-name: production
   annotations:
-    application: "Big Book of Acronyms"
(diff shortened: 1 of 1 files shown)


SteveMarshall:
@alkar Should I prefix the `namespace.yaml` with `00-`? What's the logic behind that in other namespace definitions?

alkar:
@SteveMarshall yes please - we apply with `kubectl apply -f <dir>` which will grab files in alphabetical order. If the namespace isn't prefixed, some resources will fail on the first attempt.

alkar:
We now have a way to generate a namespace and all required resources [by use of templates](https://ministryofjustice.github.io/cloud-platform-user-docs/01-getting-started/002-env-create/#create-your-namespace-and-namespace-resources). You might want to update this PR with the latest changes.

SteveMarshall:
@alkar Thanks! I'll start this again when I've got more time to sort it.

PR #150, opened 2018-09-11
Defines NetworkPolicies for the default namespace.

alkar:
Kuberos runs in the default namespace so we need to allow ingress from nginx-ingress.


kalbir:
LGTM

PR #154, opened 2018-09-11
Add env for support labelling webhook

moj-cloud-platform-test:
The [support labelling webhook](https://github.com/ministryofjustice/cloud-platform-support-labelling-webhook) currently runs on Heroku. We are investigating running it on Kubernetes instead. 

This namespace is to try deploying the app on the cluster and seeing how well it works.

### DIFF
--- a/namespaces/cloud-platform-live-0.k8s.integration.dsd.io/support-labelling-webhook-dev/00-namespace.yaml
+++ b/namespaces/cloud-platform-live-0.k8s.integration.dsd.io/support-labelling-webhook-dev/00-namespace.yaml
@@ -1,0 +1,4 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: support-labelling-webhook-dev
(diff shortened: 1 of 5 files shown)


namespaces/cloud-platform-live-0.k8s.integration.dsd.io/support-labelling-webhook-dev/00-namespace.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: support-labelling-webhook-dev
+  labels:
+    name: support-labelling-webhook
+    cloud-platform.justice.gov.uk/business-unit: cloud-platform
+    cloud-platform.justice.gov.uk/is-production: false
```
alkar: This might break, I believe it has to be a string, `"false"`, unless kubernetes automatically converts it.

kalbir: Okay let me change that. I did read somewhere that it was generally safer to put all the values as strings unless you explicitly wanted them to be e.g. numbers or whatever.

PR #158, opened 2018-09-12
Add elasticache to formbuilder-platform-dev namespace

aldavidson:
Also refactored some of the required tags into variables.tf to avoid duplication


namespaces/cloud-platform-live-0.k8s.integration.dsd.io/formbuilder-platform-dev/resources/main.tf
```diff
@@ -33,3 +36,34 @@ resource "kubernetes_secret" "rds-instance" {
...
+}
+
+resource "kubernetes_secret" "publisher-elasticache" {
+  metadata {
+    name      = "elasticache-formbuilder-publisher-dev"
+    namespace = "formbuilder-platform-dev"
+  }
+
+  data {
+    endpoint          = "${module.publisher-elasticache.endpoint}"
```
alkar: `endpoint` is not an output of the module.
You should instead use `cache_nodes`, however, it's a map so you can't put it in a secret as is. For a single-node EC instance, you can use something like [this](https://github.com/ministryofjustice/cloud-platform-terraform-elasticache-cluster/blob/master/example/outputs.tf#L7) or set them as separate fields in the secret.

At some point during the changes, we did have `endpoint` as an output, perhaps we missed some part in the docs when updating?

namespaces/cloud-platform-live-0.k8s.integration.dsd.io/formbuilder-platform-dev/resources/variables.tf
```diff
@@ -0,0 +1,9 @@
+variable "team_name"                  { default = "formbuilder-dev" }
```
alkar: Since these values do not change, you could use [`locals`](https://www.terraform.io/docs/configuration/locals.html). I'm not asking you to change them, just FYI :)

namespaces/cloud-platform-live-0.k8s.integration.dsd.io/formbuilder-platform-dev/resources/main.tf
aldavidson: I probably either misread it, or was working off my own memory of a previous version. It's been quite a day today... :)
Fix pushed

aldavidson:
@alkar i've made the changes you requested - thanks!

PR #166, opened 2018-09-18
Terraform files to automate creation kubernetes manifest files.

jojuolape:
As part of the process to automate creation of kubernetes manifest files. This PR changes the manifest values to include variables for the parts that will be created dynamically by terraform.

**Why**
Users will have to create kubernetes manifest values to deploy their app to kubernetes

***What***
This PR includes Terraform files that will automate the process of creating these Kubernetes manifest files. All users will have to do is enter the values and this will dynamically create the kubernetes value files.


namespace-resources/variables.tf
```diff
@@ -0,0 +1,11 @@
+variable "cluster" {
+    description = "What cluster are you deploying your namespace. I.E cloud-platform-test-1 "
```
alkar: Should we have a default value for `live-0` here?

namespace-resources/variables.tf
```diff
@@ -0,0 +1,11 @@
...
+}
+
+variable "namespace" {
+    description = "Namespace you would like to create on cluster <application>-<environment>. I.E myapp-dev"
+}
+
+variable "github_team" {
+    description = "Your github team"
```
alkar: I can see an issue here. Team name actually has to be the name as defined in the github API. For example, it seems it's all lowercase and spaces translated to dashes. You can get the proper name if you visit the team page and look at the URL. How do we let the user know what is the correct value?

namespace-resources/variables.tf
```diff
@@ -0,0 +1,11 @@
+variable "cluster" {
+  description = "What cluster are you deploying your namespace. I.E cloud-platform-test-1 "
+}
+
+variable "namespace" {
+  description = "Namespace you would like to create on cluster <application>-<environment>. I.E myapp-dev"
+}
+
+variable "github_team" {
+  description = "Your github team"
```
alkar: I can see an issue here. Team name actually has to be the name as defined in the github API. For example, it seems it's all lowercase and spaces translated to dashes. You can get the proper name if you visit the team page and look at the URL. How do we let the user know what is the correct value?

namespace-resources/variables.tf
```diff
@@ -0,0 +1,11 @@
+variable "cluster" {
+  description = "What cluster are you deploying your namespace. I.E cloud-platform-test-1 "
```
alkar: Should we have a default value for live-0 here?

namespace-resources/main.tf
```diff
@@ -0,0 +1,69 @@
...
+  template = "${file("${path.module}/03-resourcequota.yaml")}"
+
+  vars {
+    namespace = "${var.namespace}"
+  }
+}
+
+resource "local_file" "03-resourcequota" {
+  content  = "${data.template_file.resourcequota.rendered}"
+  filename = "../namespaces/${local.cluster}/${var.namespace}/03-resourcequota.yaml"
```
alkar: I wonder if it would be better to prepend these paths with `${path.module}` as well.

alkar:
Additionally, can we please incorporate the labels & annotations from [here](https://github.com/ministryofjustice/cloud-platform/issues/205#issuecomment-417338122)?

namespace-resources/variables.tf
jojuolape: Changed

namespace-resources/variables.tf
jojuolape: changed description

namespace-resources/main.tf
jojuolape: discussed.

alkar:
As discussed, this LGTM now. I'll confirm the actual intention for the two annotations and return with feedback.

PR #168, opened 2018-09-19
Set up Tiller in jenny-myapp-dev namespace

jennyd:
I've put this in `01-rbac.yaml` to keep all access-related config together, rather than creating a separate file for it as the docs suggest.


jennyd:
The build failed on this, but it looks like the CircleCI config is new and still being set up on a branch so merging anyway.

alkar:
@jennyd that's correct, apologies for the confusion.

PR #178, opened 2018-09-27
CircleCI Terraform validate and format tests

nayoa:
connects to ministryofjustice/cloud-platform#293

**WHAT**
- CircleCI test to check if Terraform code is validated and formatted 
- Terraform fmt kuberos `ecr.tf`and formbuilder `main.tf`
- CircleCI status to README.md

**WHY**
To catch Terraform code that is not valid before approved and merged

Note: This ticket was originally created from user feedback to catch issues with manifests before approved. After a disucssion with @alkar, using `yamllint` did not achieve this objective and requires further investigation


alkar:
@nayoa can you please squash and organise the commits before merging? Thanks

PR #180, opened 2018-10-01
add service accounts for the FormBuilder user datastore

aldavidson:


namespaces/cloud-platform-live-0.k8s.integration.dsd.io/formbuilder-services-staging/formbuilder-services-staging-admin-role.yaml
```diff
@@ -9,12 +9,18 @@ subjects:
...
   - kind: ServiceAccount
     name: formbuilder-publisher-workers-dev
     namespace: formbuilder-platform-dev
   - kind: ServiceAccount
     name: formbuilder-publisher-workers-staging
     namespace: formbuilder-platform-staging
+  # ...but only the staging user datastore can read the staging
+  # service tokens
+  - kind: ServiceAccount
+    name: formbuilder-user-datastore-staging
```
alkar: Here you've omitted the `namespace` attribute. I believe it will default to the current namespace but seeing that you're being explicit in dev you might want to add it here as well?

aldavidson: good catch - now fixed

aldavidson: oops, adding that commit means it's dismissed your review - could you re-approve please?

PR #208, opened 2018-10-11
Add UAT env to Apply LAA service

lostie:
**NOTE:** Configuration settings currently the same as for the `staging` environment.


lostie:
@alkar I've updated the configurations using the `terraform` gen. Could you please 👀 ? Thx.

PR #210, opened 2018-10-11
Add kubewatch namespace to Apply LAA 

lostie:
⚠️ ⚠️ ⚠️ **DO NOT MERGE** ⚠️ ⚠️ ⚠️ 

To deploy [kubewatch](https://github.com/bitnami-labs/kubewatch) so that we can monitor k8s deploys, etc.

**QUESTIONS:**

* According to the setup instructions we need to run something like this in order to deploy it:

```
helm install kubewatch stable/kubewatch --set='rbac.create=true,slack.channel=#YOUR_CHANNEL,slack.token=xoxb-YOUR_TOKEN,resourcesToWatch.pod=true,resourcesToWatch.daemonset=true'
```

(or use a configuration file). 

**Where are we supposed to keep these configurations? Are we expected to create a MOJ github repo just to hold this?**

### DIFF
--- a/namespaces/cloud-platform-live-0.k8s.integration.dsd.io/laa-apply-for-legalaid-kubewatch/00-namespace.yaml
+++ b/namespaces/cloud-platform-live-0.k8s.integration.dsd.io/laa-apply-for-legalaid-kubewatch/00-namespace.yaml
@@ -1,0 +1,12 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: laa-apply-for-legalaid-kubewatch
+  labels:
+    cloud-platform.justice.gov.uk/is-production: "false"
+    cloud-platform.justice.gov.uk/environment-name: "production"
+  annotations:
+    cloud-platform.justice.gov.uk/business-unit: "laa"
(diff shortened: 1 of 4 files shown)


namespaces/cloud-platform-live-0.k8s.integration.dsd.io/kubewatch/01-rbac.yaml
```diff
@@ -0,0 +1,13 @@
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: kubewatch-admin
+  namespace: kubewatch
+subjects:
+  - kind: Group
+    name: "github:apply-for-legalaid"
```
lostie: **NOTE:** Not sure if this is correct. This service could be generic across the MOJ?

nayoa: @lostie This is correct as you will be responsible for kubewatch in your namespace

lostie:
Closing. As discussed offline with @alkar, we decided to implemented an alternative solution using Prometheus Alerts.

PR #219, opened 2018-10-16
Add multi-env reference-app deployment

jasonBirchall:
**WHAT**
1. Added environments for dev, staging, and production for the reference application. 
2. Created a CircleCI service account in new namespaces. 
3. Deleted the old `reference-app` namespace.

**WHY**
1. This will allow me to demonstrate using a multi-stage muti-env pipeline. 
2. Creating a service account for CircleCI will allow me to create my pipeline for deployment. 
3. Avoids any confusion.


namespaces/cloud-platform-live-0.k8s.integration.dsd.io/reference-app-development/00-namespace.yaml
```diff
@@ -1,12 +1,12 @@
...
-  name: reference-app
+  name: reference-app-development
   labels:
     cloud-platform.justice.gov.uk/is-production: "false"
     cloud-platform.justice.gov.uk/environment-name: "development"
   annotations:
-    cloud-platform.justice.gov.uk/business-unit: "cloud-platform"
-    cloud-platform.justice.gov.uk/application: "Reference App"
-    cloud-platform.justice.gov.uk/owner: "Cloud Platform: platforms@digital.justice.gov.uk"
+    cloud-platform.justice.gov.uk/business-unit: "digital"
```
kalbir: So it turns out we have both done this wrong:

screen shot 2018-10-17 at 13 16 38

I think that means the right answer is:

```suggestion
    cloud-platform.justice.gov.uk/business-unit: "HQ"
```

PR #231, opened 2018-10-23
Add c100-application-staging resources

zheileman:


namespaces/cloud-platform-live-0.k8s.integration.dsd.io/c100-application-staging/resources/variables.tf
```diff
@@ -0,0 +1,40 @@
+variable "team_name" {
```
kalbir: This is a really good approach, I think we'll probably update the docs to recommend doing it like this. 

However, at the moment this won't work as the file needs to be called `terrafrom.tfvars` to be automatically picket up by `terraform apply`. Any other filename needs to be specified as a command line input and we don't support that at the moment. See here for some more info: https://www.terraform.io/intro/getting-started/variables.html#from-a-file

So if you rename this file to `terraform.tfvars` it should work.

zheileman: But it says in the docs the file can be named anything as long as it has the `.tf` extension, since Terraform loads all files ending in .tf in a directory:

https://www.terraform.io/intro/getting-started/variables.html#defining-variables

Other teams are naming it also `variables.tf`
I'm confused 😄

kalbir: @zheileman Aha I see what's happened. Terraform will load any `.tf` file and process it. So it is doing it with these files that are called `variables.tf`. However, terraform _also_ recommends that you put variables in files with the extension `.tfvars`, which are not all loaded, they are loaded on a case by case basis (ie. when you tell terraform to load it). I'm guessing that is to give you control over when you load different variables (e.g. if you have a `dev.tfvars` and a `prod.tfvars`, you can specify which one you need on an occasion).

I think there is a case for using `.tfvars` but seeing as everyone is using `variables.tf` at the moment I won't make you change it 😄 We'll work out which way works best for us and then come back if we need to change it.

zheileman: Yep, happy to follow the conventions and best practises, so I will happily change it in the future if needed 👍

PR #260, opened 2018-11-26
Monitoring network policy and ServiceMonitor

ollieanwyll:


namespaces/cloud-platform-live-0.k8s.integration.dsd.io/logging/04-networkpolicy.yaml
```diff
@@ -25,3 +25,21 @@ spec:
...
+metadata:
+  name: allow-prometheus-scraping
+  namespace: logging
+spec:
+  ingress:
+  - from:
+    - namespaceSelector:
+        matchLabels:
+          component: monitoring
+  podSelector:
```
alkar: On a second thought, I think it's better to set this to `podSelector: {}` since the value of the `app` label can differ. Same for the other cluster.

PR #268, opened 2018-11-29
Added new namespace for cica

andydtaylor77:
We are in the process of using the cloud platform to create a namespace for the cica environment practice.


namespaces/cloud-platform-live-0.k8s.integration.dsd.io/cica-namespace-dev-practice/01-rbac.yaml
```diff
@@ -0,0 +1,13 @@
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: cica-namespace-dev-practice-admin
+  namespace: cica-namespace-dev-practice
+subjects:
+  - kind: Group
+    name: "github:https://github.com/Adesanluj/hello-world.git"
```
jasonBirchall: @andydtaylor77 This isn't quite right. The value we're looking for is `github:<teamName>` for example, I noticed you're in the `CICA` team, which would mean your value would be `github:CICA`.

andydtaylor77: Hi @jasonBirchall this is just a proof of concept to get us started. We will be creating a proper development with more meaningful variable values after this. If it is essential to update this one I will get it changed. Thanks Andy

andydtaylor77: Hi @jasonBirchall I have made this and a couple of other changes and re-submitted. Hope this will be ok. Also did the same for an S3 bucket.

PR #295, opened 2018-12-13
aws region duplicate

razvan-moj:


namespaces/cloud-platform-live-0.k8s.integration.dsd.io/offender-management-staging/resources/rds-allocations-api.tf
```diff
@@ -2,10 +2,6 @@ terraform {
   backend "s3" {}
 }
```
alkar: Probably want to remove this bit as well?

PR #301, opened 2018-12-13
apply only to modified namespaces

razvan-moj:
**Why**
A full tf apply takes 8 mins, this will filter only dirs modified by the last commit

### DIFF
--- a/bin/apply
+++ b/bin/apply
@@ -13,6 +13,8 @@ log() {
   echo -e "${_fg}>>> ${@}\033[0m"
 }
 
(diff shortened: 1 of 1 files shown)


alkar:
This changes the behaviour. It no longer ensures that version controlled resources stay true to the source (which would be git).

We discussed having this run on pushes and forcing a full run when the pipeline is triggered by cron, is that not possible?

Also, I thought we agreed that this is not a problem yet and that we wouldn't change it just yet?

razvan-moj:
> We discussed having this run on pushes and forcing a full run when the pipeline is triggered by cron, is that not possible?

Don't know yet, will leave as do-not-merge until I figure out a way

razvan-moj:
will probably be refactored in a different manner

PR #313, opened 2018-12-21
Increase the memory given to moj-s3-virus-scan to 1Gi

misaka:


alkar:
@misaka this will change the *default* memory limit for any `Pod` in that namespace. Is that what you want to do? Perhaps it would be preferable to override the values directly on your `Deployment` (which you don't need to PR for).

That being said, if you have a single kind of service running in this namespace, then relying on the `LimitRange` could work.

alkar:
It's defined per-container, see [this example](https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/#meaning-of-memory).

alkar:
Closing this due to inactivity. Feel free to re-open or reach out to us directly.

PR #317, opened 2019-01-03
Remove jenny-myapp-dev namespace

jennyd:
I don't need this any more.


SteveMarshall:
Who needs to merge this? @jennyd?

When this is merged will that automatically remove the namespaces, too?

jennyd:
Last time I removed a namespace I was asked not to merge it immediately because there was a manual step that needed to be done first.

alkar:
@SteveMarshall we do not automatically remove namespaces at this point while we refine the relevant scripts (staying on the safe side).

@jennyd that is correct, previously there was a risk that we would lose track of deleted namespaces and associated resources. We've improved our internal process since then so feel free to merge PRs like this in the future.

PR #323, opened 2019-01-09
Add rds to laa apply for legalaid staging and production

mpw5:
Add /resources/rds.tf to staging and production namespaces to create RDS databases for the service.


mpw5:
Thanks @razvan-moj. I've updated to v3.0

PR #329, opened 2019-01-14
Add Elasticache to LAA apply for legalaid for UAT, Staging and Production

Kalagan:


Kalagan:
👍 Thanks for the quick review

PR #330, opened 2019-01-15
Increase resource limit for OMIC prod & staging

KPobeeNorris:
We have been experiencing an issue on staging where when we are trying
to deploy new code only 2 out of the 3 pods are successfully updated,
and means that we currently have to manually delete the 3rd pod to get
everything working properly.  Therefore we need to increase our
resource limit to avoid this issue moving forward.


razvan-moj:
I'll apply this as Concourse isn't up yet

PR #337, opened 2019-01-15
Fix team_names for publisher-test and platform-test-dev namespaces

solidgoldpig:


solidgoldpig:
After discussion with Al Davidson, it seems that this suffix-ation of the team_name is an unnecessary artefact, so I'm binning this change.

PR #340, opened 2019-01-16
Update names of two ECR repo secrets

jennyd:
This will make it clearer what the secrets are for in the context of the
other secrets in our namespace.


jennyd:
Only one of these files contains this - do both or neither need it?

```
terraform {
  backend "s3" {}
}

provider "aws" {
  region = "eu-west-1"
}
```

razvan-moj:
> Only one of these files contains this - do both or neither need it?

Those values need to be defined once, and only once. You could have a 'main.tf' to include them separate from other modules just for clarity.

jennyd:
Thanks @razvan-moj, I'll move them out to a `main.tf` in a separate PR 👍

PR #345, opened 2019-01-18
Adjust LimitRange and ResourceQuota for a number of namespaces

alkar:
The previous default values for new namespaces were causing scheduling issues and were not properly thought out.
This PR introduces new defaults that should work for the majority of apps we host and should address the
overprovisioning of limits.


sldblog:
📝 To capture our discussion offline: these numbers don't mean much in the end, so I'm happy to leave them as-is.

PR #356, opened 2019-01-22
Add configmaps access to circleci role for laa-cla-public namespaces

said-moj:
This is required so that circleci is able to push up configmaps for the grafana dashboards


said-moj:
Thank you both for your feedback, I have updated the PR now.

PR #360, opened 2019-01-23
Add missing LICENSE file

vijay-veeranki-moj:
Add missing LICENSE in cloud-platform-environments repo


lukaszraczylo:
I'm not certainly sure if presence of the license file is required as it's technically not a software but configuration files. Worth discussing though.

PR #361, opened 2019-01-24
LGA-288: Reduce default cla-public production limits

sldblog:
Our monitoring identified that our maximum memory usage for these pods was 126MiB since Christmas Eve 2018.

Similarly, our highest CPU spike was 59 millicpu.

We propose to scale memory down to 512Mi and CPU down to 250m. This will give us 24 pods to scale up, if necessary.


sldblog:
🎩

PR #369, opened 2019-01-29
Creating namespace and resources for vv-sprint-deploy-test

vijay-veeranki-moj:
This is to create a Cloud Platform Environment as part of deploy an app using the guide for this sprint starting 28th Jan

### DIFF
--- a/namespaces/cloud-platform-live-0.k8s.integration.dsd.io/vv-sprint-deploy-test/00-namespace.yaml
+++ b/namespaces/cloud-platform-live-0.k8s.integration.dsd.io/vv-sprint-deploy-test/00-namespace.yaml
@@ -1,0 +1,12 @@
+apiVersion: v1
+kind: Namespace
+metadata:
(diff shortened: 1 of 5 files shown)


namespaces/cloud-platform-live-0.k8s.integration.dsd.io/vv-sprint-deploy-test/00-namespace.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: vv-sprint-deploy-test
+  labels:
+    cloud-platform.justice.gov.uk/is-production: "false"
+    cloud-platform.justice.gov.uk/environment-name: "dev"
+  annotations:
+    cloud-platform.justice.gov.uk/business-unit: "cloud-platfomr"
```
ollieanwyll: Don't think it will be an issue, but a minor typo here

PR #387, opened 2019-02-04
Created file 'plan' to run terraform plan when PR is raised

vijay-veeranki-moj:
Created file 'plan' to run terraform plan when PR is raised


vijay-veeranki-moj:
Plan_Env-Terraform-Plan

PR #412, opened 2019-02-08
Run plan on changes only

vijay-veeranki-moj:
Changes to run Terraform plan on changes only


bin/plan
```diff
@@ -13,26 +13,26 @@ log() {
   echo -e "${_fg}>>> ${@}\033[0m"
 }
 
+changed_dirs=$(git diff HEAD^^..HEAD --name-only --pretty=format:"" | grep 'cloud-platform-live-0.k8s.integration.dsd.io' | awk -F'/' '{print $3}' | sort | uniq)
```
alkar: I think we'd want to compare `HEAD` to `origin/master` and not the last couple commits.

Additionally, this greps on the cluster name, shouldn't we instead filter for `resources/` for actual terraform changes?

vijay-veeranki-moj: fixed it considering the review comments.

PR #448, opened 2019-02-21
Fix conflicting definition for Parliamentary Questions ECR

lukaszraczylo:


namespaces/cloud-platform-live-0.k8s.integration.dsd.io/pq-staging/resources/ecr.tf
```diff
@@ -1,31 +0,0 @@
-#########################################
-# Parliamentary Questions ECR (STAGING) #
-#########################################
-
-terraform {
```
alkar: You will need to keep the `terraform` and aws `provider` blocks or this won't run at all.

PR #450, opened 2019-02-22
increase memory cpu quotas for a container.

Mjwillis:


alkar:
FWIW, it's preferable (and a better practice) to define resources for each container (specifically your `Deployment` spec or other). This only changes the default values for any `Pod` running in your namespace that does not define its own resource requirements.

Additionally, if you are going to depend on the `LimitRange` defaults, you might want to adjust the values a little bit so that the ratios of `ResourceQuota` to `LimitRange` values remain constant, otherwise you're essentially wasting part of your quota, eg:

`ResourceQuota.limits.cpu / LimitRange.default.cpu = 16.6667`, that's 16 `Pods`, however,
`ResourceQuota.limits.memory / LimitRange.default.memory = 12`, so you will be wasting that extra CPU capacity.

PR #460, opened 2019-02-26
Create separate ECR repos for Prison Visits Staff and Prison Visits Public

TobyRet:
- Separate docker repos created for each application
- Removes creation of RDS instance as we will be pointing to an existing db


razvan-moj:
do note this will delete and recreate the ECRs

PR #480, opened 2019-03-04
Create Elasticache instance for offender-management

rossjones:
Adds elasticache to both the staging and production environment for the offender-management-allocation service.


rossjones:
Thanks @razvan-moj

PR #485, opened 2019-03-05
Added check-my-diary-dev vars

sgsollie:


namespace-resources/variables.tf
```diff
@@ -5,32 +5,42 @@ variable "cluster" {
```
alkar: Please do not commit changes in `namespace-resources`!

digitalronin: To clarify what Mitch said, the stuff you've changed in `namespace-resources/variables.tf` should be in a new file; `namespaces/cloud-platform-live-0.k8s.integration.dsd.io/check-my-diary-dev/resources/variables.tf` (or, you could just fill in the values directly in any `namespaces/cloud-platform-live-0.k8s.integration.dsd.io/check-my-diary-dev/resources/*.tf` files you create.

sgsollie: - yeah sorry I actually wasn't supposed to push that. 

I'll copy that out elsewhere

sgsollie:
Thanks guys - I was just keeping the variables handy for later, that .tf file itself probably isn't really needed to actually proceed.

PR #494, opened 2019-03-07
Add rds resources

Adesanluj:


mtrblmoj:
Hi,  
There is no need for the aws provider stanza in rds-postgres.tf, it has already been defined elsewhere; the issue should go away after you remove it.

PR #496, opened 2019-03-07
Amended the source code URL.

njpreddy:


digitalronin:
I've raised a PR to fix the user guide wrt. https vs ssh to git clone the demo app.

PR #499, opened 2019-03-07
Preprod for CatTool

Mjwillis:


Mjwillis:
@digitalronin it is a production environment, preprod is basically a cut of production data from nomis - 1 week old - so data is sensitive.  We will then build a prod environment once PEN testing and assurance are complete.  I have update the S3 bucket to production

PR #506, opened 2019-03-12
Rename elasticache terraform file - the extension should be '.tf'

TobyRet:


razvan-moj:
2 errors: 
`terraform fmt`
and
`Error: Error asking for user input: Error parsing address 'kubernetes_secret.ec-cluster-offender-management-allocation-manager-${var.environment-name}': invalid resource address "kubernetes_secret.ec-cluster-offender-management-allocation-manager-${var.environment-name}"
`
so the var doesn't belong there

PR #513, opened 2019-03-15
Add new postgresql role

zheileman:
The intended result of this, is to create a new role (postgres user, with ability to login to the DB) with separate credentials (but still randomly generated) to the main user.

This will expose a new secret with the connection string for this new user.


zheileman:
@alkar @razvan-moj This is a draft PR, still working out some bits, but does it make sense to you?
Concourse is not happy tho, `Error initializing PostgreSQL client`.
This is in relation to the conversation we had the other day on slack.

There is however the question of how I'm going to demote privileges to this new user as the terraform provider doesn't have any capabilities to issue GRANT or other commands as far as I can tell.

zheileman:
The database and the pipeline runs in different VPC so, closing PR as I investigate a different approach.

PR #519, opened 2019-03-18
Added RDS instance to check-my-diary-dev

sgsollie:


alkar:
@sgsollie this does not apply cleanly:

```
* aws_db_instance.rds: Error creating DB Instance: InvalidParameterValue: DBName must begin with a letter and contain only alphanumeric characters.
```
See under `DBName` [here](https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_CreateDBInstance.html) for a complete set of restrictions.

Alternatively, you can leave the db name property empty and it will auto-generate one for you.

Can you please amend? Thank you!

PR #524, opened 2019-03-19
live-0: Define kiam namespace

alkar:


digitalronin:
Just so we have the context in this thread, @alkar please could you add a note explaining why this shouldn't be merged yet?

alkar:
@digitalronin this is simply because it's not been tested: we have not done this for live-0 (the namespace is managed by helm) and could potentially break `kiam` due to the `NetworkPolicies`.

jasonBirchall:
After a chat with Mitch we decided testing this was a much lower priority. As such, we will be closing this PR.

PR #535, opened 2019-03-26
Parametrise configuration for the plan script

alkar:
This will allow us to use this same script to drive pipelines for any cluster independently.

Connects to https://github.com/ministryofjustice/cloud-platform/issues/666

### DIFF
--- a/bin/.common.sh
+++ b/bin/.common.sh
@@ -1,0 +1,31 @@
+# shellcheck shell=sh
+
+log() {
+  _fg=''
+  [ "${1}" = "red" ] && _fg='\033[0;31m'
+  [ "${1}" = "blue" ] && _fg='\033[0;34m'
+  [ "${1}" = "green" ] && _fg='\033[0;32m'
+  shift
+  # shellcheck disable=SC2145
+  echo -e "${_fg}>>> ${@}\033[0m"
+}
+
+_checkenv(){
+  missing=""
+  for v in \
+      PIPELINE_CLUSTER \
+      PIPELINE_STATE_BUCKET \
+      PIPELINE_STATE_KEY_PREFIX \
+      PIPELINE_CLUSTER_STATE_BUCKET \
+      PIPELINE_CLUSTER_STATE_KEY_PREFIX; do
+    fv=""
+    eval fv="\${$v+xyz}"
+    if [ -z "${fv}" ]; then
+      missing="${missing}\n\t${v}"
+    fi
+  done
+  if [ ! -z "${missing}" ]; then
+    echo -e "missing environment variables:${missing}"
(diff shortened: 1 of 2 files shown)


bin/.common.sh
```diff
@@ -0,0 +1,31 @@
...
+  _fg=''
+  [ "${1}" = "red" ] && _fg='\033[0;31m'
+  [ "${1}" = "blue" ] && _fg='\033[0;34m'
+  [ "${1}" = "green" ] && _fg='\033[0;32m'
+  shift
+  # shellcheck disable=SC2145
+  echo -e "${_fg}>>> ${@}\033[0m"
+}
+
+_checkenv(){
```
digitalronin: I think you could avoid needing _checkenv if you `set -euo pipefail` at the top of your script.

More info here: http://redsymbol.net/articles/unofficial-bash-strict-mode/

bin/plan
```diff
@@ -1,42 +1,40 @@
...
+      log blue "Planning terraform resources for namespace ${namespace} in ${cluster}"
+      terraform init \
+        -backend-config="bucket=${PIPELINE_STATE_BUCKET}" \
+        -backend-config="key=${PIPELINE_STATE_KEY_PREFIX}${cluster}/${namespace}/terraform.tfstate" \
+        -backend-config="region=${PIPELINE_STATE_REGION}"
+      terraform plan \
+        -var="cluster_name=${cluster%%.*}" \
+        -var="cluster_state_bucket=${PIPELINE_CLUSTER_STATE_BUCKET}" \
+        -var="cluster_state_key=${PIPELINE_CLUSTER_STATE_KEY_PREFIX}${cluster%%.*}/terraform.tfstate"
+      cd $OLDPWD
```
digitalronin: Where is OLDPWD being set?

Also, you can avoid needing to do that kind of thing by wrapping your stuff in () to invoke a sub-shell, e.g.

```
(
    cd foo
    do some stuff
)
```

After this, the script will continue with pwd being whatever it was before the (

bin/.common.sh
alkar: Absolutely, the reason `_checkenv` was used in the first place is to get a more useful message, saying "you need *all* these variables set".

Happy to change, it will reduce complexity.

bin/plan
alkar: `OLDPWD` is a shell variable set by `cd` - I'm not sure whether it's an official POSIX shell variable (although it seems to be implemented) but is definitely part of bash https://www.tldp.org/LDP/abs/html/internalvariables.html

Using a sub-shell is much nicer (and we actually use it above!) indeed, will make the adjustment .

PR #550, opened 2019-03-28
Increase limits.memory quota for laa-apply-for-legalaid-uat

Kalagan:
Each PR our team create boots its own UAT env and so we are struggling to stay within the quota at the moment


alkar:
@Kalagan may I suggest we have a proper look into the resource allocation & usage for this namespace? Given its dynamic nature, perhaps we could find a better way to deal with this instead of occasionally bumping the quota.

alkar:
LGTM!

Kalagan:
@alkar, Yes we can try to find another way to fix this problem!

PR #554, opened 2019-03-28
Added tiller rbac for check-my-diary

sgsollie:


namespaces/cloud-platform-live-0.k8s.integration.dsd.io/check-my-diary-dev/01-rbac.yaml
```diff
@@ -14,3 +14,23 @@ roleRef:
...
+metadata:
+  name: tiller
+  namespace: check-my-diary-dev
+subjects:
+- kind: ServiceAccount
+  name: tiller
+  namespace: check-my-diary-dev
+roleRef:
+  kind: ClusterRole
+  name: cluster-admin
```
alkar: Can you please change this to `admin`? The permissions should be equivalent, but `admin` allows us to define them centrally them as we introduce new features / changes.

PR #555, opened 2019-03-28
Remove live-1 MTP resources because they were created in the wrong AWS region

ushkarev:


ushkarev:
@digitalronin will the ECR and RDS I removed here be deleted automatically in time? I want to make another PR to make new ones and I can still access the ECR from the cli (the secret for RDS never appeared in my namespace so don't know if it was ever made and/or deleted).

digitalronin:
@ushkarev No, the AWS resources will need to be deleted manually. 

I'm not sure what we can do about the RDS, but if you supply the ECR details, we can delete that for you.

More info here: https://user-guide.cloud-platform.service.justice.gov.uk/deploying-applications.html#cleaning-up

PR #570, opened 2019-04-01
Remove MTP live-0 namespaces and resources

ushkarev:
…this frees an ECR and an RDS instance in eu-west-1


ushkarev:
see slack comments on PR for AWS resource details

PR #597, opened 2019-04-05
Moved cccd-dev from live-0 to live-1

colinbruce:


colinbruce:
Thanks @digitalronin, I'll update those.  Is there a method of obtaining/list of the latest versions for the terraform modules, or should I just be looking at the templates?

alkar:
@colinbruce you'll need to check the repositories individually I'm afraid. There is a [list](https://github.com/ministryofjustice/cloud-platform#terraform-modules) of the available ones for easy reference.

PR #600, opened 2019-04-05
Update the ecr credentials for cccd-dev

colinbruce:
We are trying to standardise tour internal naming convention and these had diverged


colinbruce:
That's fine... it's probably empty as I've been trying to push to `laa-get-paid/cccd` anyway! :)

PR #617, opened 2019-04-10
migrate formbuilder-product-page-prod to live-1

asmega:
- this is a copy paste of the existing `formbuilder-product-page-prod`
- i have not created the ECR repo yet

### DIFF
--- a/namespaces/live-1.cloud-platform.service.justice.gov.uk/formbuilder-product-page-prod/00-namespace.yaml
+++ b/namespaces/live-1.cloud-platform.service.justice.gov.uk/formbuilder-product-page-prod/00-namespace.yaml
@@ -1,0 +1,12 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: formbuilder-product-page-prod
+  labels:
+    cloud-platform.justice.gov.uk/is-production: "true"
+    cloud-platform.justice.gov.uk/environment-name: "prod"
+  annotations:
+    cloud-platform.justice.gov.uk/business-unit: "transformed-department"
+    cloud-platform.justice.gov.uk/application: "Formbuilder Product Page"
+    cloud-platform.justice.gov.uk/owner: "Form Builder: formbuilder@digital.justice.gov.uk"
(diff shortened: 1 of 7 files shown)


namespaces/live-1.cloud-platform.service.justice.gov.uk/formbuilder-product-page-prod/00-namespace.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: formbuilder-product-page-prod
+  labels:
+    cloud-platform.justice.gov.uk/is-production: "false"
```
digitalronin: Should "is-production" be "true" here?

namespaces/live-1.cloud-platform.service.justice.gov.uk/formbuilder-product-page-prod/resources/ecr.tf
```diff
@@ -0,0 +1,27 @@
+terraform {
+  backend "s3" {}
+}
+
+provider "aws" {
+  region = "eu-west-2"
+}
+
```
digitalronin: I'd recommend extracting lines 1-8 into a separate main.tf file

namespaces/live-1.cloud-platform.service.justice.gov.uk/formbuilder-product-page-prod/resources/ecr.tf
```diff
@@ -0,0 +1,27 @@
...
+module "formbuilder_product_page_ecr_credentials" {
+  source    = "github.com/ministryofjustice/cloud-platform-terraform-ecr-credentials?ref=2.1"
```
digitalronin: In live-1 you can (and should) use the latest version of the module - currently "ref=3.1"

asmega:
@digitalronin thanks for the feedback. I have made all the suggested changes. and yes that value should be true for the production flag

PR #619, opened 2019-04-11
Add ecr to new laa-legal-adviser-api namespace

exonian:
This ecr.tf is just a copy of the one that's in the staging namespace on live-0

I know it won't need to go in more than one namespace dir in the end, but I thought it best to add it to this dir as we're going to be cleaning the live-0 one up soon, but apologies if I've misunderstood.

### DIFF
--- a/namespaces/live-1.cloud-platform.service.justice.gov.uk/laa-legal-adviser-api-staging/resources/ecr.tf
+++ b/namespaces/live-1.cloud-platform.service.justice.gov.uk/laa-legal-adviser-api-staging/resources/ecr.tf
@@ -1,0 +1,27 @@
+terraform {
+  backend "s3" {}
+}
+
+provider "aws" {
+  region = "eu-west-1"
+}
+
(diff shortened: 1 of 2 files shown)


namespaces/live-1.cloud-platform.service.justice.gov.uk/laa-legal-adviser-api-staging/resources/ecr.tf
```diff
@@ -0,0 +1,27 @@
+terraform {
+  backend "s3" {}
+}
+
+provider "aws" {
+  region = "eu-west-1"
+}
+
+module "ecr-repo-api" {
+  source = "github.com/ministryofjustice/cloud-platform-terraform-ecr-credentials?ref=2.1"
```
digitalronin: Please update this to use the latest version of the module (i.e. `ref=3.1`) You can find this by visiting the repo URL and checking for the latest entry in `tags`

PR #622, opened 2019-04-11
rm live-0 formbuilder-product-page-prod resources

asmega:
- We have migrated the app to `live-1` and cleaning up `live-0`
- Delete `live-0` `formbuilder-product-page-prod` resources
- No `main.tf` file existed so I have created one


asmega:
thank you

PR #628, opened 2019-04-15
Add Pingdom support

Mjwillis:


namespaces/cloud-platform-live-0.k8s.integration.dsd.io/categorisation-tool-production/resources/pingdom.tf
```diff
@@ -0,0 +1,17 @@
+
+resource "pingdom_check" "cloud-platform-prometheus-live-0-healthcheck" {
```
jasonBirchall: Hi. Can you change `cloud-platform-prometheus-live-0-healthcheck`, it shouldn't really make a difference other than readability.

Mjwillis: done @jasonBirchall

PR #638, opened 2019-04-16
Modify ecr.tf  add main.tf and variable.tf

ermiasworku:


ermiasworku:
Further change required.

PR #683, opened 2019-04-25
add wildcard certificates to formbuilder envs

asmega:


asmega:
fyi when this get merged there will need to be a manual step in AWS. i believe to setup the dns zones?

asmega:
hmm merging is still blocked. is it because you are using a different github account?

asmega:
that worked. thank you!

razvan-moj:
> hmm merging is still blocked. is it because you are using a different github account?

yes, was in the wrong browser

PR #714, opened 2019-05-01
nomis-api-access-staging environment

mattops:


namespaces/live-1.cloud-platform.service.justice.gov.uk/nomis-api-access-staging/resources/main.tf
```diff
@@ -0,0 +1,15 @@
+terraform {
+  backend "s3" {}
+}
+
```
digitalronin: Please reinstate the default AWS provider here. Otherwise the pipeline will fail for modules which don't use the 'aws.london' provider.

PR #728, opened 2019-05-08
update variable.tf

ermiasworku:


digitalronin:
Actually, it doesn't matter - that namespace has been deleted, so it's best to submit a new PR that simply deletes all the files in this namespace sub-directory.

PR #733, opened 2019-05-09
Set up Prison Visits Staging on Live-1 Cluster

KPobeeNorris:


namespaces/live-1.cloud-platform.service.justice.gov.uk/prison-visits-booking-staging/resources/elasticache.tf
```diff
@@ -0,0 +1,23 @@
+module "ec-cluster-prison-visits-booking-staff" {
+  source                 = "github.com/ministryofjustice/cloud-platform-terraform-ecr-credentials?ref=3.3"
```
alkar: I think you need to use `github.com/ministryofjustice/cloud-platform-terraform-elasticache-cluster?ref=3.0` here.

KPobeeNorris: Thanks @alkar - will update now.

PR #735, opened 2019-05-09
Update money-to-prisoners namespaces with latest terraform resource versions

ushkarev:
and disable ECR lifecycle policy
c.f. [Issue 839](https://github.com/ministryofjustice/cloud-platform/issues/839)

### DIFF
--- a/namespaces/live-1.cloud-platform.service.justice.gov.uk/money-to-prisoners-prod/resources/ecr.tf
+++ b/namespaces/live-1.cloud-platform.service.justice.gov.uk/money-to-prisoners-prod/resources/ecr.tf
@@ -1,10 +1,15 @@
 # https://github.com/ministryofjustice/cloud-platform-terraform-ecr-credentials
 
 module "ecr" {
-  source = "github.com/ministryofjustice/cloud-platform-terraform-ecr-credentials?ref=3.1"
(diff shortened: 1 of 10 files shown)


namespaces/live-1.cloud-platform.service.justice.gov.uk/money-to-prisoners-prod/resources/ecr.tf
```diff
@@ -1,10 +1,15 @@
...
 
-  team_name = "${var.team_name}"
-  repo_name = "${var.application}"
+  providers = {
+    aws = "aws.london"
+  }
+
+  team_name     = "${var.team_name}"
+  repo_name     = "${var.application}"
+  enable_policy = false
```
vijay-veeranki-moj: No need to mention this , please remove  "enable_policy = false"
As setting this will remove the policy applied in the module.

ushkarev: but that's entirely the idea we discussed!

vijay-veeranki-moj: The policy is defined in the module and set default = true, and the limit is set to 100.
As you are using the module defined, and when you set enable_policy = false, that will remove the policy applied.

PR #745, opened 2019-05-13
Updated circleci config

vijay-veeranki-moj:
This is to run the Terraform validate and format on just the changed resources

related to #826

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,7 +13,7 @@ jobs:
       - run:
           name: Validate Terraform configurations
           command: |
-            changed_resources=$(git diff --no-commit-id --name-only HEAD..master | grep /resources/ | sort | uniq)
+            changed_resources=$(git diff --no-commit-id --name-only -r origin/master...HEAD | grep /resources/ | sort | uniq)
             for _f in ${changed_resources}; do
               find ${_f} -exec dirname {} \;|sort -u | while read m; 
                 do 
@@ -23,7 +23,7 @@ jobs:
       - run:
           name: Check if Terraform configurations are properly formatted
           command: |
-            changed_resources=$(git diff --no-commit-id --name-only HEAD..master | grep /resources/ | sort | uniq)
+            changed_resources=$(git diff --no-commit-id --name-only -r origin/master...HEAD | grep /resources/ | sort | uniq)
             for _f in ${changed_resources}; do
               tf=$(terraform fmt -write=false -diff ${_f}); 
               if [[ -n "$tf" ]]; then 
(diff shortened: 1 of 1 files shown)


vijay-veeranki-moj:
> Just curious, what's the difference between `find ${_f} -exec dirname {} \;` and `dirname ${_f}` here? Isn't `${_f}` always a single filename, at this point?

Thanks @digitalronin agree with you, fixed it

.circleci/config.yml
```diff
@@ -12,15 +12,29 @@ jobs:
       - checkout
       - run:
           name: Validate Terraform configurations
-          command: find . -type f -name "*.tf" -exec dirname {} \;|sort -u | while read m; do (terraform validate -check-variables=false "$m" && echo "√ $m") || exit 1 ; done
+          command: |
+            changed_resources=$(git diff --no-commit-id --name-only -r origin/master...HEAD | grep /resources/ | sort | uniq) || true
+            for _f in ${changed_resources}; do
+              dirname ${_f} | while read m;
```
alkar: Is this `while` loop needed?
Would it work if we simply had the following?
```
(terraform validate -check-variables=false "$(dirname ${_f})" && echo "√ $(dirname ${_f})") || exit 1 ;
```

.circleci/config.yml
```diff
@@ -12,15 +12,29 @@ jobs:
...
+                done 
+            done 
       - run:
           name: Check if Terraform configurations are properly formatted
-          command: tf=$(terraform fmt -write=false -diff); if [[ -n "$tf" ]]; then printf "$tf\n\n\nSome terraform files failed syntax validation, run 'terraform fmt' to fix\n\n"; exit 1; fi
+          command: |
+            changed_resources=$(git diff --no-commit-id --name-only -r origin/master...HEAD | grep /resources/ | sort | uniq) || true
+            for _f in ${changed_resources}; do
+              tf=$(terraform fmt -write=false -diff ${_f}); 
+              if [[ -n "$tf" ]]; then
```
alkar: To reduce complexity, we could treat this like the other check above. From `terraform fmt -help`:
```
  -check=false     Check if the input is formatted. Exit status will be 0 if all input is properly formatted and non-zero otherwise.
```
We can use that (perhaps with `-diff` as well, in order to show the actual problem) and end up with something like:
```
(terraform fmt -check -diff ${_f}) || exit 1;
```
That would make this snippet very similar to the validate snippet above.

.circleci/config.yml
vijay-veeranki-moj: Thanks Mitch agree with your suggestion, updated

.circleci/config.yml
vijay-veeranki-moj: agree with this as well to keep it simple and similar as above, updated

PR #751, opened 2019-05-13
Remove PVB RDS instance

KPobeeNorris:
We need to remove the current RDS instance from PVB entirely as a work
around for bumping the Postgres version.  A subsequent PR will be
created to re-add the RDS instance with the 9.6 version.


KPobeeNorris:
Though that might be easier as I'll be adding them back in shortly :)

KPobeeNorris:
Adding them as previously suggested

PR #756, opened 2019-05-14
Rename keys in laalaa rds secret data

exonian:
Given that they're within an `rds` secret, these data keys make more sense to us when consuming them during our deployment


exonian:
Sorry, realised after submitting this earlier that we might want to change the name of the secret itself, so held off merging for that

vijay-veeranki-moj:
> Sorry, realised after submitting this earlier that we might want to change the name of the secret itself, so held off merging for that

@exonian This will delete existing secrets and create new ones, hope its ok?

exonian:
> @exonian This will delete existing secrets and create new ones, hope its ok?

@vijay-veeranki-moj Yep absolutely fine, thanks for checking

PR #762, opened 2019-05-17
Remove amq-broker to unblock pipeline till issue is fixed

said-moj:
Remove amq-broker to unblock pipeline till issue is fixed


sldblog:
> unblock pipeline till issue is fixed

Could you provide some context on what happened, please?

sldblog:
Copying from Slack:

> We're not 100% sure what the specific issue is, but we narrowed the issue down to a lack of permission on the concourse side.

PR #780, opened 2019-05-23
Adding amq integration with laa-legal-adviser-api

said-moj:
Adding amq integration with laa-legal-adviser-api

### DIFF
--- a/namespaces/live-1.cloud-platform.service.justice.gov.uk/laa-legal-adviser-api-staging/resources/main.tf
+++ b/namespaces/live-1.cloud-platform.service.justice.gov.uk/laa-legal-adviser-api-staging/resources/main.tf
@@ -4,4 +4,5 @@ terraform {
 
 provider "aws" {
   region = "eu-west-2"
+  alias  = "london"
 }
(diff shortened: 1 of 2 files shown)


mtrblmoj:
Failed because circleCI was using tf 0.12. 
We changed that version back to 0.11.11 (known working). Manually merging this.

namespaces/live-1.cloud-platform.service.justice.gov.uk/laa-legal-adviser-api-staging/resources/main.tf
```diff
@@ -5,3 +5,8 @@ terraform {
 provider "aws" {
   region = "eu-west-2"
 }
+
+provider "aws" {
```
sldblog: There are now two provider entries, will that cause issues?
https://github.com/ministryofjustice/cloud-platform-environments/blob/e52cdec13fd98cab89f5fbf5e17f59a967431f0e/namespaces/live-1.cloud-platform.service.justice.gov.uk/laa-legal-adviser-api-staging/resources/main.tf#L5-L12

sldblog: I see they've been added separately on purpose, could the reason behind that be shared, please?

mtrblmoj: Hi, 

No this won't cause any issue/conflict.

The first provider in this main.tf is the default one (no alias), and is used for almost everything. 
Providers with aliases ("aws.london") are only used when they are explicitly invoked, like for the amq module. 

We have chosen this approach to clarify which regions is used by resources/modules.

PR #782, opened 2019-05-23
two new ecr repos for probation services on eu-west-1

andymarke:


vijay-veeranki-moj:
Hi @andymarke, can you please run `terraform fmt` and fix the fmt issue

PR #784, opened 2019-05-24
Namespace for TZ cronjobber

razvan-moj:
**Why**
Requested in https://github.com/ministryofjustice/cloud-platform/issues/804


razvan-moj:
tzupdater is not needed, our nodes have /usr/share/zoneinfo/* from the Debian package

digitalronin:
Please could you add a note in the user guide explaining that this option for user cron jobs exists, and where to find the details?

namespaces/live-1.cloud-platform.service.justice.gov.uk/cronjobber/build.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
```
alkar: I think we should have this as a separate job in concourse (building the image). Right now I'm assuming the intention is to build manually whenever we need to?

razvan-moj:
> Please could you add a note in the user guide explaining that this option for user cron jobs exists, and where to find the details?

Sure, docs will be a separate PR in the other repo

namespaces/live-1.cloud-platform.service.justice.gov.uk/cronjobber/build.sh
razvan-moj: kube-system is a small change so sure, will check Concourse too and amend

PR #793, opened 2019-05-28
CCCD - rationalise environments

colinbruce:
We have been working as a team to migrate the app across.  While we have done this a couple of naming discrepancies have crept in.

Also, I made some poor variable choice names at the start! ;)

This PR aims to get the two namespaces back in sync.  All previously created artifacts are transitory and can be deleted and re-created if needed.


colinbruce:
Hmm, concourse failed with
```
module.cccd_dev_rds.provider.aws.london
``` 
but that was the old, now deleted, namespace.

digitalronin:
The problem is that you're renaming the module and upgrading it to a different version, in the same PR.

If you break those two stages into separate PRs, it should work. Please give that a try.

colinbruce:
I've removed the RDS files for now. I'll try that on a completely separate PR.

colinbruce:
Breaking everything down into smaller PR's for debugging :'(

PR #803, opened 2019-05-29
Delete unused c100-application-production live-0 namespace

zheileman:
As part of a previous (old) PR #620, we deleted any AWS resources in the `production` namespace, but we didn't remove it as we were waiting for a migration to happen.

Now we are all good and this Live-0 namespace can be deleted.
We believe we've deleted everything inside the namespace (pods, secrets, ingress, etc.).


zheileman:
@vijay-veeranki-moj Hi, CI-validate is complaining about the deleted `resources` folder.

Has the procedure to delete a namespace changed since the last time I removed one? It used to be first the AWS are removed (but not the resources folder with the main.rf) and then, as part of a second PR, the whole namespace folder is removed.

vijay-veeranki-moj:
> Has the procedure to delete a namespace changed since the last time I removed one? It used to be first the AWS are removed (but not the resources folder with the main.rf) and then, as part of a second PR, the whole namespace folder is removed.

@zheileman The error is from circleci which is trying to check the terraform fmt, as the files are removed, it coudn't find them to do terraform fmt.

Procedure not been changed

Please ignore that error we need to fix the tf fmt check.

zheileman:
@vijay-veeranki-moj cool thanks. Will merge this then 👍

PR #815, opened 2019-05-31
Add region to s3 secret

exonian:
So that django-storages in our app can work out which version of
signature to use


exonian:
Sorry @razvan-moj I hadn't realised we had this resource on our prod namespace already so had to go back and make the change there too

PR #817, opened 2019-05-31
Part 2 of removing court-probation-dev from live0

andymarke:


andymarke:
I'm guessing `Validate Terraform configurations` build step can't cope with the whole tree of directories being deleted?

.circleci/config.yml
```diff
@@ -13,14 +13,14 @@ jobs:
       - run:
           name: Validate Terraform configurations
           command: |
-            changed_resources=$(git diff --no-commit-id --name-only -r origin/master...HEAD | grep /resources/ | sort | uniq) || true
+            changed_resources=$(git diff --diff-filter=d --no-commit-id --name-only -r origin/master...HEAD | grep /resources/ | sort | uniq) || true
```
andymarke: filter deleted files else `terraform fmt` will fail trying to validate a file that no longer exists.

PR #824, opened 2019-06-04
Create four sqs queues for cccd-dev namespace

kmahern:
Create 4 queues for the cccd-dev namespace using ref=2.0


colinbruce:
It would be good to have a meeting with someone from the Cloud Platforms team about this PR. Having four sets of credentials to monitor a set of queues managed by the same system seems a bit over the top!  It would be good to discuss the set up in conjunction with the issue I raised here https://github.com/ministryofjustice/cloud-platform-terraform-sns-topic/issues/9

PR #825, opened 2019-06-04
Expand namespace to another Get Access team

sldblog:
The @ministryofjustice/laa-get-access-data-and-reporting team has been created recently and the resources in this namespace will be shared between these two teams.

The intention of this pull request is to assign the admin role to both of these GitHub teams.


sldblog:
🎩 🙇

PR #914, opened 2019-06-19
Removing pecs namespace as requested by owner

jasonBirchall:
There was a problem that prevented the owner from achieving this


digitalronin:
@jasonBirchall can this be merged?

PR #938, opened 2019-06-26
Checkmydiary

nickb834:
Create new prod and preprod namespaces for checkmydiary using same ECR repos etc


nickb834:
Closed - think the build failed on consul as I put in two namespaces at the same time (for the same app across two envs)

PR #948, opened 2019-07-02
add ECR for form builder fb-builder

asmega:
this docker image is used to build form builder components
this is used by our continuous integration tools


asmega:
thanks!

PR #955, opened 2019-07-02
add namespace formbuilder-grafana-dashboard

asmega:
- this namespace will be used merely for a grafana dashboard
- given the way we have our code split out into several namespaces there is no logical place to call home for our grafana dashboard so I have created this namespace to house it
- it will only attach one config map which will contain the grafana config
- also a service account for circleci so any changes can be auto deployed


asmega:
I have now removed the terraform files

asmega:
actually i've changed my mind and will stick monitoring here as well

PR #956, opened 2019-07-02
add ECR for formbuilder init containers

asmega:
these docker images are used to initialise runner docker images


asmega:
Thanks David! No worries, it wasn't urgent.

PR #973, opened 2019-07-10
modify rds.tf to update providers and module version

ermiasworku:


vijay-veeranki-moj:
@ermiasworku This change will remove the existing RDS and create a new one, is that ok?

ermiasworku:
@vijay-veeranki-moj  That is fine.

PR #980, opened 2019-07-11
DT-97 Add Offender Case Note Environment

Mjwillis:


namespaces/live-1.cloud-platform.service.justice.gov.uk/offender-case-notes-dev/resources/rds.tf
```diff
@@ -0,0 +1,36 @@
+variable "cluster_name" {}
+
+variable "cluster_state_bucket" {}
+
+module "dps_rds" {
+  source                 = "github.com/ministryofjustice/cloud-platform-terraform-rds-instance?ref=4.4"
```
digitalronin: Please use the 4.5 version of this module, and add the database family value:

https://github.com/ministryofjustice/cloud-platform-terraform-rds-instance/releases/tag/4.5

Mjwillis: Done

PR #995, opened 2019-07-16
Increase redis size

mtrblmoj:
- Elasticache module 3.0 -> 3.1
- new provider block


digitalronin:
@mtrblmoj Can/should this be merged now?

digitalronin:
This change may cause a short service outage, so it needs to be scheduled with the service team.

PR #1002, opened 2019-07-18
Add disclosure-checker-production namespace

tmckernan:


zheileman:
@tmckernan we already created the ECR repo on the first namespace (staging) so here we don't need the ecr.tf file.

PR #1003, opened 2019-07-19
Add `hmpps-book-secure-move-api-production` environment

stevehook:
Based on the existing `hmpps-book-secure-move-api-staging` environment. Does not include `ecr.tf` as I believe we can use the existing ECR repo from `hmpps-book-secure-move-api-staging`.


namespaces/live-1.cloud-platform.service.justice.gov.uk/hmpps-book-secure-move-api-production/resources/rds.tf
```diff
@@ -0,0 +1,24 @@
+module "rds-instance" {
+  source = "github.com/ministryofjustice/cloud-platform-terraform-rds-instance?ref=4.0"
```
vijay-veeranki-moj: Can you use the latest version please `4.5`
for guidance
`https://github.com/ministryofjustice/cloud-platform-terraform-rds-instance/blob/master/example/rds.tf`

stevehook: thanks @vijay-veeranki-moj - have updated the version

PR #1012, opened 2019-07-23
Reduce default request values for new namespaces

digitalronin:
Averaged across the whole cluster, containers use
8m CPU, and 128Mi Memory. 

The defaultRequest values are the resources the
scheduler will *reserve* for every container in a 
pod. If these values are much higher than what
containers will actually use, the cluster will run
out of capacity much sooner than it should.

NB: This commit only changes the *request* value,
not the limits on how much resource a container
is actually allowed to consume.


digitalronin:
Sure. I'll amend the PR

PR #1034, opened 2019-07-30
Formbuilder resourcing

asmega:


asmega:
i think i follow. I have removed limits from the resourcequotas from the core formbuilder repos

digitalronin:
If you prefer, we can try applying some of the changes manually too, before committing the changes for the pipeline.

asmega:
sure. i've removed the `live` deletions. there is one `live` change but that's a change not a removal

PR #1037, opened 2019-07-30
Add jason-lab RDS instance with variables

jasonBirchall:


namespaces/live-1.cloud-platform.service.justice.gov.uk/jason-lab/resources/variables.tf
```diff
@@ -0,0 +1,79 @@
...
+variable "team_name" {
+  default = "cloud-platform"
+}
+
+variable "application" {
+  default = "jason-lab-test"
+}
+
+variable "is-production" {
+  default = "fals"
```
digitalronin: should be "false" not fals

PR #1043, opened 2019-07-31
Add route53.tf file for apply for legal aid 

willc-work:
Add a route53.tf file to set up a domain name for the production environment under the ```.service.justice.gov.uk``` domain.

The preferred names are:

production  ```apply-for-legal-aid.service.justice.gov.uk```

staging ```staging.apply-for-legal-aid.service.justice.gov.uk```

Would it be better to set the route53.tf on production and have a subdomain for staging?

### DIFF
--- a/namespaces/live-1.cloud-platform.service.justice.gov.uk/laa-apply-for-legalaid-staging/resources/route53.tf
+++ b/namespaces/live-1.cloud-platform.service.justice.gov.uk/laa-apply-for-legalaid-staging/resources/route53.tf
@@ -1,0 +1,23 @@
+resource "aws_route53_zone" "apply_for_legal_aid_route53_zone" {
+  name = "staging.apply-for-legal-aid.service.justice.gov.uk"
+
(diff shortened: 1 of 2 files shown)


namespaces/live-1.cloud-platform.service.justice.gov.uk/laa-apply-for-legalaid-staging/resources/route53.tf
```diff
@@ -0,0 +1,23 @@
+resource "aws_route53_zone" "apply_for_legal_aid_route53_zone" {
+  name = "staging.apply-for-legal-aid.service.justice.gov.uk"
```
vijay-veeranki-moj: Can you please create a zone some thing like `apply-for-legal-aid.service.justice.gov.uk`
Can see more guidance about creating a Zone [here][https://user-guide.cloud-platform.service.justice.gov.uk/tasks.html#defining-the-dns-zone]

PR #1073, opened 2019-08-07
This PR adds several improvements to the namespace reporter script

digitalronin:
* Treat the supplied name as a pattern (e.g. get output for all the prison-visits namespaces at once)
* Add a `-o json` option for JSON output
* Handle memory values in the format 99999Ki
* Handle namespaces where the limits are not set (e.g. 'default')
* Update the usage tips at the top of the file

Note: This is a breaking change, in the sense that `./bin/namespace-reporter.rb foo` will not
work anymore. Now, it has to be `./bin/namespace-reporter.rb -n foo`

### DIFF
--- a/bin/namespace-reporter.rb
+++ b/bin/namespace-reporter.rb
@@ -189,6 +189,22 @@ class Namespace
   end
 end
 
+def text_output(ns)
+  puts
+  puts "Namespace: #{ns[:name]}"
+  puts
+  puts "  Request limit:\tCPU: #{ns[:max_requests][:cpu]},\tMemory: #{ns[:max_requests][:memory]}"
+  puts "  Requested:\t\tCPU: #{ns[:resources_requested][:cpu]},\tMemory: #{ns[:resources_requested][:memory]}"
+  puts
+  puts "  Num. containers:\t#{ns[:container_count]}"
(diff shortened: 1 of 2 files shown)


bin/namespace-reporter.rb
```diff
@@ -182,27 +191,54 @@ def memory_value(str)
...
-name = ARGV.shift
+def parse_options
+  options = { format: TEXT_OUTPUT }
+
+  OptionParser.new do |opts|
+    opts.on("-n", "--namespace NAMESPACE", "Namespace name pattern (required)") do |ns|
+      options[:namespace] = ns
+    end
+
+    opts.on("-o", "--output FORMAT", "Output format (#{JSON_OUTPUT} | #{TEXT_OUTPUT})") do |fmt|
```
jsugarman: @digitalronin hi, this is great. was going to suggest `optparse` before I saw this. Am trying it out atm. one thing you could add to enforce the format options and default to text is;
```
    opts.on("-o", "--output [FORMAT]", [JSON_OUTPUT, TEXT_OUTPUT], "Output format (#{JSON_OUTPUT} | #{TEXT_OUTPUT})") do |fmt|
      raise "Unknown output format #{fmt}" unless [JSON_OUTPUT, TEXT_OUTPUT].include?(fmt)
      options[:format] = fmt || 'text'
    end
```

digitalronin: Thanks for that, @jsugarman I've updated the script with your change.

PR #1075, opened 2019-08-08
increase preprod offender manager CPU limits

starswan:


namespaces/live-1.cloud-platform.service.justice.gov.uk/offender-management-preprod/02-limitrange.yaml
```diff
@@ -6,9 +6,9 @@ metadata:
 spec:
   limits:
   - default:
-      cpu: 50m
+      cpu: 300m
       memory: 750Mi
     defaultRequest:
-      cpu: 10m
+      cpu: 50m
```
digitalronin: This change won't do anything useful. All it does is force the cluster to reserve extra capacity for every container in your pod. Your containers are allowed to *use* up to 300m each - this just makes it harder to *schedule* them.
I strongly recommend leaving your defaultRequest.cpu value at 10m
See this article for more info: https://user-guide.cloud-platform.service.justice.gov.uk/concepts.html#namespace-container-resource-limits
FYI, all the containers in this namespace are using 1m/2m of CPU right now, and between 27Mi (for metrics) and 257Mi (manager and sidekiq) of memory, according to `kubectl -n offender-management-preprod top pods --containers`
You can also override the defaultRequest values in your deployment, for each container.

starswan: reset as requested

PR #1083, opened 2019-08-13
Reduce limitrange and resourcequota for fee calc

digitalronin:
* remove namespace limits
* reduce namespace request limits
* reduce limitrange cpu request limit
* increase limitrange memory limit
* reduce limitrange cpu limit


jsugarman:
👍 looks good @digitalronin. thanks

digitalronin:
This change needs to be applied via the following series of steps: 
• change the limitrange values
• cycle all the pods
• change the resourcequota values
If you reduce the resourcequota first, you can end up with a namespace which is overcommitted (i.e. existing pods have reserved more resources than the namespace is allowed).
Once this change is approved. I'll pause the pipeline and manually apply the changes before merging and unpausing.

digitalronin:
I've applied these changes via kubectl edit and a manual cycling of all the pods, in both namespaces, so merging now.

PR #1085, opened 2019-08-13
Adjust CCCD namespace resources

digitalronin:
reduce limitrange request.cpu
increase limitrange request.memory
reduce CPU namespace request.limit


jsugarman:
👍 looks good too @digitalronin. many thanks. Will redeploy (apply to all namespaces) once merged and run through pipeline.

digitalronin:
This change needs to be applied via the following series of steps: 
• change the limitrange values
• cycle all the pods
• change the resourcequota values
If you reduce the resourcequota first, you can end up with a namespace which is overcommitted (i.e. existing pods have reserved more resources than the namespace is allowed).
Once this change is approved. I'll pause the pipeline and manually apply the changes before merging and unpausing.

digitalronin:
I've applied the limitrange changes and cycled all the pods in these namespaces

PR #1087, opened 2019-08-13
Create risk profiler database for cat tool - postgresql resource bug workaround

steverendell:


digitalronin:
@steverendell Please run `terraform fmt` and commit the update.

PR #1114, opened 2019-08-16
Remove custom Prometheus alert from PVB & MOIC

KPobeeNorris:
The KubePodNotReady alert is not working as intended and is causing a
huge amount of noise in our alerts channels for PVB & MOIC so we are
going to remove it.

### DIFF
--- a/namespaces/live-1.cloud-platform.service.justice.gov.uk/offender-management-preprod/prometheus-custom-alerts-moic-preprod.yaml
+++ b/namespaces/live-1.cloud-platform.service.justice.gov.uk/offender-management-preprod/prometheus-custom-alerts-moic-preprod.yaml
@@ -7,14 +7,6 @@ spec:
   groups:
   - name: moic-preprod
     rules:
-    - alert: KubePodNotReady
-      annotations:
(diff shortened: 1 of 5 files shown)


KPobeeNorris:
@digitalronin - our initial goal with this alert was getting it to fire if all all the containers failed to be ready after 5 mins, e.g. in PVB we have pods with three containers and thought it would be a useful alert if one/any of them weren't ready.  However, this alert only appears to look at the state of the pod and as soon as one container is ready then it shows as running.  

The noise is being generated as it finds pods for cron jobs that have completed and alerts us as they're not running, even though we added the regex to ignore any pods that are in a running or completed state.  

Any suggestions/advice/help gratefully received as always, just in it's current state it's so noisy we can't see 'real' issues when they come into the channel :)

digitalronin:
> it finds pods for cron jobs that have completed and alerts us as they're not running
🤦‍♂ That's really good to know, @KPobeeNorris Thanks for telling us.

KPobeeNorris:
@digitalronin - no worries! Going to merge for now, and look to add that alert again in the future if a fix is found :)

PR #1115, opened 2019-08-16
initial commit of Risk profiler change queue. Dev environment

lbennett-moj:


namespaces/live-1.cloud-platform.service.justice.gov.uk/offender-categorisation-dev/resources/messaging.tf
```diff
@@ -0,0 +1,69 @@
...
+  {
+    "Version": "2012-10-17",
+    "Id": "${module.risk_profiler_change.sqs_arn}/SQSDefaultPolicy",
+    "Statement":
+      [
+        {
+          "Effect": "Allow",
+          "Principal": {"AWS": "*"},
+          "Resource": "${module.risk_profiler_change.sqs_arn}",
+          "Action": "SQS:SendMessage"
```
digitalronin: I think you might need to add a condition here, as per this other SQS definition:
https://github.com/ministryofjustice/cloud-platform-environments/blob/master/namespaces/live-1.cloud-platform.service.justice.gov.uk/cccd-staging/resources/messaging.tf#L95

lbennett-moj: Thanks @digitalronin. I did look at the other example, they are writing to the queue from a sns topic - which needed a bit more work. I don't think I need a condition here as I just need to let the service send to the queue.  Would it be possible to merge so I could test?  Probably need to iterate a few times

lbennett-moj: I'll also contact the people you suggested - thanks

PR #1127, opened 2019-08-19
Delete the namespace sid-test01app-dev

digitalronin:
This namespace has no cluster or AWS resources,
and was created by sid234 on 10th June 2019.

After this change is merged, the namespace
should be manually deleted.


digitalronin:
I've deleted the namespace from the live-1 cluster

PR #1150, opened 2019-08-22
Amend limitranges back to original

jsugarman:
For further consideration by vijay
and dsalgado


jsugarman:
may need deletion of pods sol leaving for now on advice off CP

jsugarman:
Thanks @digitalronin. away on hols for couple of weeks now so will do as you suggest when i get back, unless someone on the team picks it up first.

PR #1156, opened 2019-08-27
namespace for keyworker app prod environment

mattops:


namespaces/live-1.cloud-platform.service.justice.gov.uk/manage-key-workers-prod/00-namespace.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: manage-key-workers-prod
+  labels:
+    cloud-platform.justice.gov.uk/is-production: "true"
+    cloud-platform.justice.gov.uk/environment-name: "prod"
+  annotations:
+    cloud-platform.justice.gov.uk/business-unit: "HMPPS"
```
mattops: Regarding the request to prefix the namespace with `hmpps`.  Is this annotation not enough?  We've not prefixed any of our other namespaces so to start doing so now will be a bit odd.

mattops: I've fixed the limits and quotas as requested.

digitalronin:
The naming of this service, and its domain, were discussed in this slack thread;

https://mojdt.slack.com/archives/C57UPMZLY/p1566992777005000

PR #1177, opened 2019-09-05
Create RDS instance for HMCTS Complaints Adapter

emileswarts:


namespaces/live-1.cloud-platform.service.justice.gov.uk/hmcts-complaints-formbuilder-adapter-production/resources/rds.tf
```diff
@@ -0,0 +1,30 @@
...
+  cluster_state_bucket       = "${var.cluster_state_bucket}"
+  db_backup_retention_period = "${var.db_backup_retention_period_hmcts_complaints_adapter}"
+  application                = "hmcts-complaints-formbuilder-adapter"
+  environment-name           = "${var.environment-name}"
+  is-production              = "${var.is-production}"
+  infrastructure-support     = "${var.infrastructure-support}"
+  team_name                  = "${var.team_name}"
+  force_ssl                  = true
+  db_engine_version          = "11.4"
+
```
mtrblmoj: Since this is a production RDS, let's make it bigger. Could you add : 
`db_instance_class = "db.m4.xlarge"`

emileswarts: Hi @mtrblmoj 

Thanks for the suggestion.
This database won't be heavily used, will only have 1 table and the data won't be persisted for a very long time.  

What is the default size if not specified?

mtrblmoj: 👍

digitalronin:
@emileswarts Sorry, my bad, with db_engine_version 11.4, you also need to set `rds_family = "postgres11"`

Line 32 of the module defaults it to postgres10, which is not compatible with the db_engine_version value, so it needs to be overridden.

Please could you submit a new PR with that change?

PR #1182, opened 2019-09-06
Replace Amazon MQ with Amazon ElastiCache for Redis for laa-legal-adviser-api

sldblog:
We never had any success with getting Amazon MQ working with our combination of Django and Celery.

We tried CloudAMQP for a few months, but that developed mysterious connection problems.

Recently we tried with a freebie Heroku Redis instance, and it worked like a charm.

This change replaces the Amazon MQ instance with a Redis instance (through Amazon ElastiCache) to finalise the move to Redis.

PR #1184, opened 2019-09-06
Remove environment postfix from the secret

sldblog:
A quick follow-up to #1182: the secret postfix is unnecessary as we already have "staging" in the namespace.


sldblog:
:bowing_man: :tophat: (Sorry)

PR #1186, opened 2019-09-06
Revert "Reduce "Apply for legal aid" namespaces CPU allocation"

scruti:
Reverts ministryofjustice/cloud-platform-environments#1168

We were actually allocating more CPU per pod than the Namespace Allocation we set up, so started getting alerts like: 

"Alert: Namespace laa-apply-for-legalaid-production is using 170% of its requests.cpu "

Reverting the reduction, will adjust accordingly and re-open the PR.


scruti:
Thanks @razvan-moj

PR #1192, opened 2019-09-10
Track a query poc ecr (CTS Namespace)

mseedat-moj:
As discussed with Mourad - ECR terraform config file added for Track a Query


mseedat-moj:
Discard

PR #1211, opened 2019-09-12
Reduce resources allocated to umar-dev namespace

digitalronin:
I missed this one, in #1203


jennyd:
Umar has left MOJ so this namespace can be deleted.

PR #1226, opened 2019-09-16
Add get verb to MTP Test for circleci service account

marcofucci:
This is because if you want to "patch deployments" you need "get" as well.


marcofucci:
@digitalronin you were right 😞 
Could you please have a look when you have 5 mins? I hope it's the only extra verb needed.

PR #1227, opened 2019-09-16
Add labels instead of annotation - opa pod toleration

poornima-moj:


poornima-moj:
Wrong commit with conflicts. Opening new PR

PR #1253, opened 2019-09-25
remove formbuilder-services-pentest-dev

digitalronin:
The formbuilder team have said they no longer need any of their pentest namespaces.

This commit removes formbuilder-services-pentest-dev

After this PR is merged, the namespace should be manually deleted from the cluster

`kubectl delete namespace formbuilder-services-pentest-dev`


digitalronin:
I've manually deleted this namespace from live-1

PR #1254, opened 2019-09-25
remove formbuilder-services-pentest-production

digitalronin:
The formbuilder team have said they no longer need any of their pentest
namespaces.

This commit removes formbuilder-services-pentest-production

After this PR is merged, the namespace should be manually deleted from
the cluster

`kubectl delete namespace formbuilder-services-pentest-production`


digitalronin:
I've manually deleted this namespace.

PR #1259, opened 2019-09-25
remove formbuilder-publisher-pentest namespace

digitalronin:
The formbuilder team have said they no longer need
 any of their pentest namespaces.

This commit removes 
formbuilder-publisher-pentest

After this PR is merged, the namespace should be 
manually deleted from the cluster, and a 
subsequent PR raised to remove the directory and 
the main.tf file

`kubectl delete namespace formbuilder-publisher-pentest`


digitalronin:
I've run `terraform destroy` to delete the AWS resources from this namespace

digitalronin:
I've deleted this namespace from live-1

PR #1271, opened 2019-10-01
Reduce container requests to actual usage values

digitalronin:
The concourse worker containers resource usage
looks like this;

      used (cpu, memory): 53        1279
      used (cpu, memory): 463       1043

The web container looks like this;

      used (cpu, memory): 19        73

This commit brings the concourse container
resource requests down to values which are close
to what is actually used.

After this PR is merged, it needs to be redeployed
via `terraform apply`


digitalronin:
This is a null change (whitespace at the end of the file).
I must have got confused with all the PRs affecting concourse. Closing this one.

PR #1273, opened 2019-10-02
Reduce namespace resources

digitalronin:
Change limitrange and resourcequota for
claim-criminal-injuries-compensation-staging

There is nothing running in this namespace.


digitalronin:
Closed by mistake

PR #1279, opened 2019-10-02
Fix malformed Route53 alias record from #1272

ushkarev:


ushkarev:
@vijay-veeranki-moj if this still doesn't work, i can try ignoring the `prod*` subdomains – they might not even be needed, i just felt safer having them the same as in the DSD account

PR #1309, opened 2019-10-08
check-financial-eligibility-production requests

digitalronin:
* CPU to 250m (double total requests)
* memory to 5000Mi (default for new namespaces)


digitalronin:
Opened by mistake. Ignore this.

PR #1339, opened 2019-10-10
Track a Query - Add new environments

mseedat-moj:
Adds all required environments for migrating the `correspondence_tool_staff` application (Track a Query)


digitalronin:
@mseedat-moj You've got "is-production" set to true in an lot of your elasticache.tf files, too.

digitalronin:
Both production and staging rds.tf have "is-production" true

digitalronin:
demo, production and staging s3.tf have is-production true

mseedat-moj:
@digitalronin - yes I wasn't sure what the repercussions of setting things to 'is production' to true was - where I thought we need to mimic production as closely as possible (e.g. demo, staging, production) I set `is-production: true` - advice welcome.

PR #1345, opened 2019-10-11
Add a github action to run terraform fmt

digitalronin:
This workflow runs terraform fmt against a new PR,
and commits any files which are changed as a
result of that.

depends on #1344


digitalronin:
The 'format-code' action failed because this PR depends on #1344 and that hasn't been merged yet.

PR #1378, opened 2019-10-16
Created namespace for Prepare a case (prod)

theDustRoom:


namespaces/live-1.cloud-platform.service.justice.gov.uk/prepare-a-case-prod/00-namespace.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: prepare-a-case-prod
+  labels:
+    cloud-platform.justice.gov.uk/is-production: "false"
```
digitalronin: Should "is-production" be "true" here?

theDustRoom: Yes, file updated now thanks.

PR #1389, opened 2019-10-17
CPU requests: offender-case-notes namespaces

digitalronin:
The offender-case-notes namespaces are requesting 3000m CPU each,
but the max. sum of all pod requests is 560m (this would drop to
about 50m if all pods were cycled).

This commit reduced the namespace CPU request to 600m, freeing up
7200m of unused CPU.


mattops:
LGTM

PR #1391, opened 2019-10-17
Reduce use of force namespaces CPU requests

digitalronin:
The use-of-force-(dev|preprod|prod) namespaces request 3000m CPU
each. The total CPU requested by all pods in those namespaces is
190m each.

This change reduces the CPU requested for each namespace from
3000m to 300m, freeing up 8100m of unused CPU.


mattops:
LGTM

PR #1393, opened 2019-10-17
Reduce fj-cait resource requests

digitalronin:
The fj-cait-(production|staging) namespaces are requesting 3000m
CPU and 6000Mi of memory.

The total requests for all pods in those namespaces is 50m CPU and
600Mi of memory.

This change reduces the requests for those namespaces to 100m CPU
and 1000Mi of memory.

This change should have no impact at all on the service. It does
not limit the resources available to the namespace, it just
changes the information provided to the scheduler in order to
help it decide where to schedule these namespaces' workloads.

This frees up 5800m CPU and 10Gi of memory which were not being
used, making them available to the cluster.


digitalronin:
I'm waiting for Jesus Laiz to have a look at this PR before merging it.

zheileman:
Happy with the changes @digitalronin @vijay-veeranki-moj 👍

PR #1400, opened 2019-10-18
Reduce CPU requests for legal adviser namespaces

digitalronin:
The max total of all pods' CPU requests in these two namespaces is 375m
(this will drop to 30m when the pods are restarted, and the new
limitrange values take effect). However, the namespaces request 3000m of
CPU each.

This change reduces the cpu requested by each namespace to 500m, freeing
up 5500m of unused CPU capacity.


digitalronin:
Awaiting feedback from Michael Blatherwick before merging.

exonian:
Looks good to us, thanks

digitalronin:
Waiting til Monday to merge, because I'm paranoid.

PR #1401, opened 2019-10-18
Reduce dps-welcome-(dev|prod) CPU requests

digitalronin:
The total of all the CPU requests for all pods in these namespaces
is 190m. This will drop to 20m after all pods are restarted.

This change reduces the namespaces' CPU requests from 3000m to
300m, freeing up 5400m of unused CPU capacity.


digitalronin:
Awaiting feedback from Matt Whittaker before merging

mattops:
LGTM

digitalronin:
Ready to merge. Leaving this til Monday because I'm paranoid.

PR #1417, opened 2019-10-22
Reduce CPU requests for offender-management

digitalronin:
The total of all the CPU requests for all pods in these namespaces is
800m in production, 600m in staging and 90m in preprod.  This will drop
to 90m in all 3 namespaces once all pods are restarted.

This change reduces the namespaces' CPU requests from 3000m to 1000m,
freeing up 4000m of unused CPU capacity. Although 1000m is much more
than should be required, I want all the namespaces to have the same
resourcequota, so that they all behave consistently wrt. scheduling
pods.

We can make a subsequent change to reduce the resourcequota further,
after all pods have been cycled and more resource can be reclaimed.


mattops:
LGTM

PR #1420, opened 2019-10-22
Reduce offender-categorisation ns CPU requests

digitalronin:
The total of all the CPU requests for all pods in these namespaces is
400m. This will drop to 60m once all pods are restarted, because some
pods are still running with the old limitrange CPU requests value of
180m rather than 10m.

This change reduces the namespaces' CPU requests from 3000m to 600m,
freeing up 7200m of unused CPU capacity.

We can make a subsequent change to reduce the resourcequota further,
after all pods have been cycled and more resource can be reclaimed.


mattops:
LGTM

PR #1430, opened 2019-10-23
Create dstest namespace

digitalronin:
This is a dev namespace, for experimenting with pod limits on
namespaces.


digitalronin:
Reopening this - I need a namespace so I can play with terraform state locking

PR #1442, opened 2019-10-24
Remove A record for legacy MTP domain

ushkarev:
to prepare it for migration to cloud platform


ushkarev:
currently, this domain is still hosted in DSD account, but can be moved once this is done

PR #1445, opened 2019-10-25
remove form builder integration resources

asmega:
this is in preparation to delete these namespaces


asmega:
this failed in concourse. I think someone needs to do some of this by hand as i'm deleting resources

PR #1459, opened 2019-10-29
Create authorized-keys-provider setup on live-1

vijay-veeranki-moj:
This is related to https://github.com/ministryofjustice/cloud-platform/issues/1141

We have authorized-keys-provider set up in live-0, this PR will create authorized-keys-provider in live-1 also creates a new s3 bucket.


namespaces/live-1.cloud-platform.service.justice.gov.uk/authorized-keys-provider/provider.yaml
```diff
@@ -0,0 +1,115 @@
...
+  selector:
+    matchLabels:
+      app: authorized-keys-provider
+  template:
+    metadata:
+      labels:
+        app: authorized-keys-provider
+    spec:
+      containers:
+      - name: alpine
```
jasonBirchall: Does the alpine image run as root by default, if so, it won't pass the psp requirements.

vijay-veeranki-moj: @jasonBirchall Thanks for identifying this, updated this to use ministryofjustice/cloud-platform-tools:1.4 and added securityContext, to run as non root user.

PR #1464, opened 2019-10-30
Reduce c100-application namespaces resource reqs.

digitalronin:
Current namespace requests: 3000m CPU, 12Gi memory
Current request totals: 60m CPU, 1.5Gi memory

This commit reduces namespace requests to 300m CPU and 4Gi memory


digitalronin:
Have been discussing with Jesus. Closing in favour of #1469

PR #1469, opened 2019-10-31
Reduce resourcequota for c100 app. staging

digitalronin:
This reduces CPU from 3000m to 300m (5 x current requests total),
and memory from 12Gi to 6Gi (4 x current requests total).

Jesus will try a deployment with maxSurge: 100% after this PR is
merged, and if there are no problems a subsequent PR will make the
same changes to the production namespace.


zheileman:
👍

PR #1471, opened 2019-10-31
Change c100-app prod ns. requests to match staging

digitalronin:
Jesus has tested deployment into the staging namespace, with the
new quota values, and everything seems to be fine. So this change
brings the production namespace into line with staging.


zheileman:
Great, thanks! 👍

PR #1473, opened 2019-11-01
Add prometheus rules to namespace

zheileman:
We had this in the deploy repo.

Adding the file to the environments repo so that rules can be deployed and re-created via the pipeline in case of cluster failure (essentially like we do with certificates).


zheileman:
@digitalronin yep, it's the exact same file, with the rules we already had in place 👍

PR #1478, opened 2019-11-04
Add cccd production DNS records

jsugarman:
### What
copies of existing template-deploy DNS records

### Why

In preparation for having GDS delegate control
for production service.gov.uk domain.


jsugarman:
Automatic formatting now. nice. Had just pushed a fix and had it rejected

jsugarman:
@vijay-veeranki-moj this is now ready. I have only added the A record as all other records in the TD `claim-crown-court-defence.service.gov.uk.` hosted zone are related to email in on way or another (sendgrid,DMARC, smtp MX record, amazon SES (Simple Email Service) - and CCCD user govuk notify nowadays.

PR #1498, opened 2019-11-06
Track a query r53 records

mtrblmoj:
When creating route53 records with terraform should not be the full name. 
Instead an A record's name should only be the prefix to be addedd to zone name (if any) 

`track-a-query.service.gov.uk.` ->  `.`
`staging.track-a-query.service.gov.uk.` -> `staging`


mtrblmoj:
This has been manually applied during testing.

PR #1515, opened 2019-11-07
probation-core-services-tooling. Adds postresql database for sonarqube

robinsillem:


namespaces/live-1.cloud-platform.service.justice.gov.uk/probation-core-services-tooling/resources/sonarqube-db.tf
```diff
@@ -0,0 +1,44 @@
...
+
+  providers = {
+    # Can be either "aws.london" or "aws.ireland"
+    aws = "aws.london"
+  }
+}
+
+resource "kubernetes_secret" "pcs_sonarqube_rds" {
+  metadata {
+    name      = "pcs_sonarqube_rds"
```
digitalronin: I think you need to replace the underscores here with hyphens.

This is the pipeline error:

```
Error: kubernetes_secret.pcs_sonarqube_rds: metadata.0.name a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')
```

robinsillem: OK, done, thanks

PR #1538, opened 2019-11-11
Replace cpu/memory limits with pod count limits

digitalronin:
related to #1534

This change alters production namespaces, which
were excluded from #1534

Nobody reported any issues with the earlier change, so it should be safe to apply this change to production namespaces. 

However, I recommend leaving it a couple of days before merging this PR, just to be safe.


digitalronin:
It's been 2 days, with no issues arising from the change to non-production namespaces, so I think it's safe to merge this now.

PR #1552, opened 2019-11-12
Track a Query - Add S3 CORS rules for staging bucket

mseedat-moj:
Track a Query uses Dropzone in the browser to allow users to upload documents directly into S3. This module amendment will set the appropriate CORS rules for both the cloud platform generated URL and the actual staging.track-a-query domain

### DIFF
--- a/namespaces/live-1.cloud-platform.service.justice.gov.uk/track-a-query-staging/resources/s3.tf
+++ b/namespaces/live-1.cloud-platform.service.justice.gov.uk/track-a-query-staging/resources/s3.tf
@@ -4,7 +4,7 @@
 #################################################################################
 
 module "track_a_query_s3" {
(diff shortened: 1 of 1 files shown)


namespaces/live-1.cloud-platform.service.justice.gov.uk/track-a-query-staging/resources/s3.tf
```diff
@@ -61,4 +61,20 @@ resource "kubernetes_secret" "track_a_query_s3" {
     bucket_arn        = "${module.track_a_query_s3.bucket_arn}"
     bucket_name       = "${module.track_a_query_s3.bucket_name}"
   }
+
+  cors_rule = [
```
mtrblmoj: Sorry I missed this. 
cors_rule should not be in the kubernetes_secret, but in the module block

PR #1555, opened 2019-11-12
Convert the namespace creation to a ruby script

digitalronin:
This script has slightly more advanced features than
the existing terraform-based process (e.g. when we
want 'true' or 'false', we keep asking until we get
one of those values), and it has a framework for
adding additional validators, e.g. regexp matching for
the source code url.

This change is being made now so that the gitops work
can build on this ruby script, rather than having a
change to ruby at the same time as changing the
functionality.


digitalronin:
Ugh. It's executable for me, even if I delete the file and restore it via `git checkout` I'll try 2 commits - one to chmod 644 and a second to chmod 755

digitalronin:
Ah, I think it's a problem with the code-formatter github action. I think that is changing the mode of files. I'll see what I can do to fix that.

PR #1560, opened 2019-11-12
Track a Query - Add production S3 bucket CORS

mseedat-moj:


mtrblmoj:
the module version needs to be bumped to 3.4

PR #1604, opened 2019-11-21
DT-363: Create case notes to probation preprod and prod environments

petergphillips:


namespaces/live-1.cloud-platform.service.justice.gov.uk/case-notes-to-probation-prod/00-namespace.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: case-notes-to-probation-prod
+  labels:
+    cloud-platform.justice.gov.uk/is-production: "false"
```
digitalronin: Should "is-production" be "true" here?

petergphillips: :man_facepalming: hadn't spotted that, now fixed (and two other of envs that didn't have the flag set either)

PR #1612, opened 2019-11-21
Adds S3 buckets to Book A Secure Move API environments

spikeheap:
This PR adds S3 buckets to our Book A Secure Move API application and stores the secrets so our app can extract them later.

Looking at existing implementations, it doesn't _appear_ to be necessary to add IAM policies as we're just using the generated credentials in our API container.

### DIFF
--- a/namespaces/live-1.cloud-platform.service.justice.gov.uk/hmpps-book-secure-move-api-preprod/resources/s3.tf
+++ b/namespaces/live-1.cloud-platform.service.justice.gov.uk/hmpps-book-secure-move-api-preprod/resources/s3.tf
@@ -1,0 +1,33 @@
+/*
+ Based on https://github.com/ministryofjustice/cloud-platform-terraform-s3-bucket/tree/master/example
+ */
+module "book_a_secure_move_documents_s3_bucket" {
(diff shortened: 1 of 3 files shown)


spikeheap:
It's not clear why the Circle build fails... but it appears to affect all the open PRs at the moment...

namespaces/live-1.cloud-platform.service.justice.gov.uk/hmpps-book-secure-move-api-preprod/resources/s3.tf
```diff
@@ -0,0 +1,33 @@
...
+
+  providers = {
+    # Can be either "aws.london" or "aws.ireland"
+    aws = "aws.london"
+  }
+}
+
+resource "kubernetes_secret" "book_a_secure_move_documents_s3_bucket" {
+  metadata {
+    name      = "book_a_secure_move_documents_s3_bucket"
```
vijay-veeranki-moj: Hi @spikeheap , plan pipeline failed, it is complaining about the name, i cannot have "_" in a name as shown in the error.
`https://concourse.cloud-platform.service.justice.gov.uk/builds/37177`

spikeheap:
I've fixed the broken Concourse test :+1:

PR #1648, opened 2019-11-26
Add dns zone and cert for new fala staging custom domain

exonian:
If this is approved, would somebody be able to test it pre-merge please?

`find-legal-advice.justice.gov.uk` is in production use on `live-0` so if these DNS changes to a subdomain of it would disrupt that, that either needs to not happen or to be rolled back asap

### DIFF
--- a/namespaces/live-1.cloud-platform.service.justice.gov.uk/laa-fala-staging/resources/route53.tf
+++ b/namespaces/live-1.cloud-platform.service.justice.gov.uk/laa-fala-staging/resources/route53.tf
@@ -1,22 +1,22 @@
 resource "aws_route53_zone" "route53_zone" {
   name = "staging.find-legal-advice.justice.gov.uk"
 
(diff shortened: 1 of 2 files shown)


namespaces/live-1.cloud-platform.service.justice.gov.uk/laa-fala-staging/resources/route53.tf
```diff
@@ -0,0 +1,22 @@
+resource "aws_route53_zone" "route53_zone" {
+  name = "staging.find-legal-advice.justice.gov.uk"
```
vijay-veeranki-moj: There is a zone exists with name `find-legal-advice.justice.gov.uk`.

 You can use a sub-domain for the existing zone, by creating an ingress using host name "staging.find-legal-advice.justice.gov.uk"

This will avoid creating multiple zones.

PR #1678, opened 2019-11-29
Update terraform RDS definitions for MTP

ushkarev:


ushkarev:
looks like the 2 failures are about terraform-0.12 not being set up on concourse and within the github action… am i right?

PR #1682, opened 2019-12-02
Added ecr.tf and adapted for namespace and team

mattempty:
Added ecr.tf and followed guidance for adapting definitions.


mattempty:
Incorrect values. Re-submitting.

PR #1713, opened 2019-12-04
Add a ruby script to delete a namespace and its AWS resources

digitalronin:
part of https://github.com/ministryofjustice/cloud-platform/issues/1437
related to https://github.com/ministryofjustice/cloud-platform/issues/1345


digitalronin:
This needs work - the tests try to actually talk to S3, which they shouldn't. I'll fix it.

digitalronin:
I've fixed the specs so that all external services are mocked - i.e. the tests don't actually talk to s3 or kubernetes.

PR #1721, opened 2019-12-05
Add a script to notify us about deleted namespaces

digitalronin:
This script is designed to be executed by a concourse pipeline, whenever a
commit to the environments repo is merged to master. It will send a notification 
to #cloud-platform if there are any namespaces which should be deleted. The 
notification includes a link to the runbook for 'manually' deleting a namespace 
and its AWS resources, using the deleter script.

This is a temporary manual step until we are confident that the only
namespaces which are being flagged for deletion are the right ones.
Subsequent commits will alter this script to delete the namespaces by
itself, and merely notify us that this has happened.


digitalronin:
Closed in favour of #1722

PR #1729, opened 2019-12-06
Track a Query - Add pingdom to demo

mseedat-moj:


digitalronin:
@mseedat-moj  Is this still required? If so, please could you merge it? If not, we'll close this PR in a few days.

PR #1739, opened 2019-12-10
terraform for new Offender search service (preprod and prod environme…

lbennett-moj:
…nts)


offender-search-prod/00-namespace.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: offender-search-prod
+  labels:
+    cloud-platform.justice.gov.uk/is-production: "false"
```
digitalronin: This namespace is called "prod", but you've got "is-production" "false" and "environment-name" "dev" - is that correct?

lbennett-moj: thanks - will address

PR #1749, opened 2019-12-16
Add a docker registry pull-through cache to the cluster

digitalronin:
A subsequent PR will alter the kops config so that worker nodes
will use this cache, when pulling images from docker hub.

Of course the cache will not be present in a new cluster, even
though the workers will be configured to use it, but this does
not cause problems - the docker daemon on the worker nodes will
try the cache first, but will fail over to contacting docker
hub directly if the cache is not responding.

see: https://github.com/ministryofjustice/cloud-platform-docker-registry-cache

related to: https://github.com/ministryofjustice/cloud-platform/issues/1443


digitalronin:
We want to restrict the cache to only be accessible from the worker nodes - not from the whole internet. So, this PR needs to wait on some more testing.

digitalronin:
Superseded by https://github.com/ministryofjustice/cloud-platform-infrastructure/pull/522

PR #1765, opened 2019-12-19
Add laa-fala-staging cert

exonian:


digitalronin:
@exonian Is this still required? If so, please could you merge it? If not, we'll close this PR in a few days.

PR #1767, opened 2019-12-23
Merge the 'gitops' development work into master

digitalronin:
The gitops-ruby branch was under active development for some time, during which
there were a lot of changes made to master.

This PR brings the gitops-ruby code up to date with respect to the current
state of master, and adds more rspec tests to both the main and the gitops
execution paths when users create namespaces and the pipeline applies changes.

Note that although there is decent unit test coverage, there may be unforeseen
problems when the pipeline runs, so I'm not planning to merge this PR until w/c
06/01/20.

co-authored with @jasonBirchall and @sablumiah 

closes https://github.com/ministryofjustice/cloud-platform/issues/1516

related to https://github.com/ministryofjustice/cloud-platform/issues/1126


bin/create-gitops-namespace-files.rb
```diff
@@ -0,0 +1,212 @@
...
+# TODO: default
+# TODO: question N of M
+# TODO: validate answers
+# TODO: check namespace name is available
+# TODO: remove old terraform stuff from namespace-resources/
+# TODO: mount user's home directory into the tools image, so we can write deployment files to their working copy of their app.
+
+TEMPLATES_DIR = "gitops-namespace-resources"
+DEPLOY_TEMPLATES_DIR = "#{TEMPLATES_DIR}/kubectl_deploy"
+CLUSTER_NAME = "gitops-test.cloud-platform.service.justice.gov.uk"
```
jasonBirchall: This will need to be live-1 for master. Thinking about it, should we set this variable in the pipeline?

spec/fixtures/gitops-deploy-files/ingress.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: helloworld-rubyapp-ingress
+  namespace: make-gitops-namespace-test-dev
+spec:
+  tls:
+  - hosts:
+    - make-gitops-namespace-test-dev.apps.gitops-test.cloud-platform.service.justice.gov.uk
```
jasonBirchall: This is needs to be a variable and will not work on clusters not named `gitops-test`

bin/create-gitops-namespace-files.rb
jasonBirchall: Fixed.

spec/fixtures/gitops-deploy-files/ingress.yaml
jasonBirchall: This file is used for comparison not execution.

PR #1787, opened 2020-01-07
Add a script to delete manually created pods

digitalronin:
This script will delete any pods which are:

  * not part of a ReplicaSet
  * and not in kube-system
  * and have been running for more than 2 days

Such pods prevent the node-recycler from draining a node to
replace it.

Developers need to run manual pods for tasks such as port-forwarding
to a database, but there's no good reason for such a pod to still be
running after 48 hours. If pods should always be running, they should
be part of a deployment (and hence a ReplicaSet).

NB: This PR adds the script. A subsequent PR will
create a concourse pipeline to execute it daily.


digitalronin:
I'm closing this PR because I've messed up the git commit history. I'll open another.

PR #1790, opened 2020-01-09
updating to helm3, no longer need tiller

mattops:


mattops:
My understanding is that this only allow permissions within the namespace it's applied to.  If I understand correctly this is `Allows read/write access to most objects in a namespace. It does not allow viewing or modifying roles or rolebindings`

PR #1859, opened 2020-01-27
GDPR-29 Adding RDS resource to the Data Compliance service (dev)

brightonsbox:


namespaces/live-1.cloud-platform.service.justice.gov.uk/prison-data-compliance-dev/resources/rds.tf
```diff
@@ -0,0 +1,33 @@
...
+  }
+}
+
+resource "kubernetes_secret" "dps_rds" {
+  metadata {
+    name      = "dps-rds-instance-output"
+    namespace = "${var.namespace}"
+  }
+
+  data {
```
mattops: think missing an equals here, ```data = {```

PR #1861, opened 2020-01-28
add new namespace for formbuilder submit product

asmega:
- i've reduced the out of the box 50 pods to 10 as it seemed overkill for this particular setup


digitalronin:
Oops - didn't notice this was a draft PR.

asmega:
not any more :)

PR #1880, opened 2020-02-03
Ensure gitops pipeline works with Manager cluster

jasonBirchall:


lib/cp_env/namespace_dir.rb
```diff
@@ -4,6 +4,7 @@ class NamespaceDir
     attr_reader :executor
 
     GITOPS_RESOURCES_DIR = "gitops-resources"
+    MANAGER_CLUSTER = "arn:aws:eks:eu-west-2:754256621582:cluster/manager"
```
digitalronin: I'd probably call this "MANAGER_CLUSTER_ARN" so that people don't misinterpret it as being the cluster name, or something.

PR #1903, opened 2020-02-11
Changed github group to github:dps-shared

GurnankCheema:
Had previously used an incorrect group name


digitalronin:
The file namespaces/live-1.cloud-platform.service.justice.gov.uk/resources/ecr.tf should be inside your namespace folder. i.e. it should be `namespaces/live-1.cloud-platform.service.justice.gov.uk/[your namespace]/resources/ecr.tf`

PR #1912, opened 2020-02-12
Adds a role that can be assumed from this namespace

robinsillem:
This is an untested best-guess from reading the kiam docs and converstaions on #ask-cloud-platform


robinsillem:
This is a PoC for connectivity between Cloud Platform and Delius

PR #1931, opened 2020-02-18
Adding laa-cla-frontend-staging namespace and ecr component

said-moj:
Was previously on live-0 which is no longer available


namespaces/live-1.cloud-platform.service.justice.gov.uk/laa-cla-frontend-staging/resources/ecr.tf
```diff
@@ -0,0 +1,23 @@
...
+  }
+}
+
+resource "kubernetes_secret" "ecr-repo" {
+  metadata {
+    name      = "ecr-repo-laa-cla-frontend"
+    namespace = var.namespace
+  }
+
+  data {
```
vijay-veeranki-moj: Can you fix this by adding 
`data = {`

PR #1968, opened 2020-02-26
Formbuilder services pentest production

mattempty:
Namespace for the pentest environment for formbuilder (5 of 5).


namespaces/live-1.cloud-platform.service.justice.gov.uk/formbuilder-services-pentest-production/00-namespace.yaml
```diff
@@ -0,0 +1,17 @@
...
+# Source: formbuilder-services/templates/00-namespace.yaml
+# auto-generated from fb-cloud-platforms-environments
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: formbuilder-services-pentest-production
+  labels:
+    # for fb's own purposes
+    name: formbuilder-services-pentest-production
+    cloud-platform.justice.gov.uk/is-production: "false"
```
jasonBirchall: You have production in the namespace name. We intend to treat this namespace as non-production if this tag is false. Can you confirm this is non-production?

mattempty: Thanks @jasonBirchall  - I am replicating the form builder environment, which was copied from the formbuilder-services-test-production namespace. So it will not be a production.

jasonBirchall: test-production??? :confused: 

Okay, as long as you understand that this namespace and all services in it would be treated as non-critical in the event of a disaster for example.

PR #1984, opened 2020-02-28
Add pingdom check to form builder

tomas-stefano:
## What this PR does?

This PR adds the pingdom check to the main apps running on formbuilder-services-* namespace.

## Context

Within the formbuilder app called fb-publisher, every time a new user create a form and publish the form, we provision 2 more new pods into formbuilder-services-* namespace, with every pod being a single tenant app.

So we decided to go for the route of using Cloud Platform pingdom integration but then it means every time an user publish a new form we need to add here the form. The solution is good for now but will be revised the moment we have more forms than we currently have at the moment.

### DIFF
--- a/namespaces/live-1.cloud-platform.service.justice.gov.uk/formbuilder-monitoring/resources/main.tf
+++ b/namespaces/live-1.cloud-platform.service.justice.gov.uk/formbuilder-monitoring/resources/main.tf
@@ -1,0 +1,13 @@
+terraform {
+  backend "s3" {
+  }
+}
+
+provider "aws" {
+  region = "eu-west-2"
+}
+
+provider "aws" {
+  alias  = "london"
+  region = "eu-west-2"
+}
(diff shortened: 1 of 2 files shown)


tomas-stefano:
@mogaal cheers thank you for pointing it out. Done! ♻️

namespaces/live-1.cloud-platform.service.justice.gov.uk/formbuilder-monitoring/resources/main.tf
```diff
@@ -0,0 +1,13 @@
...
+}
+
+provider "aws" {
+  region = "eu-west-2"
+}
+
+provider "aws" {
+  alias  = "london"
+  region = "eu-west-2"
+}
```
brenetic: @mogaal Do we need to have two definitions of the aws provider if they're both for the same region? This seems to have been done in other `main.tf` files

PR #2005, opened 2020-03-04
Update Dmarc after instructions from Tony Bishop

digitalali-moj:
Our Dmarc setting seems to be wrong from before migration, this has caught the attention of Tony Bishop who has instructed us to change it to the new literal.


poornima-moj:
@digitalali-moj , let us know when you want to merge the DMARC changes, and we will keep an eye on the route53 record update.

PR #2035, opened 2020-03-12
Add script to rotate RDS password

digitalronin:
This script will:

  * taint the random_string.password of the RDS instance
  * terraform apply to replace the password and kubernetes secret
  * gracefully cycle all the namespace pods


digitalronin:
#2040 includes changing from random_string.password to random_password

Once we've done that, this script is either redundant or, if we still need it, should update the random_password rather than the random_string.password.

Either way, it shouldn't be merged until we have everyone on release 5.3 or later of the RDS module

digitalronin:
superseded by #2258 because I screwed up the commit history

PR #2038, opened 2020-03-13
Track a Query - Added temporary RDS resource for restore of production db

Polsonby:


namespaces/live-1.cloud-platform.service.justice.gov.uk/track-a-query-production/resources/rds-restore.tf
```diff
@@ -0,0 +1,54 @@
...
+  snapshot_identifier = "rds:cloud-platform-e400332aa0770e41-2020-03-13-02-34"
+
+  providers = {
+    aws = aws.london
+  }
+}
+
+resource "kubernetes_secret" "track_a_query_rds_restore" {
+  metadata {
+    name      = "track-a-query-rds-output"
```
mtrbls: can we rename this as well please ?

PR #2047, opened 2020-03-17
applying new probation events queue to prod and removing obsolete one.

lbennett-moj:


namespaces/live-1.cloud-platform.service.justice.gov.uk/offender-events-prod/resources/pathfinder-sub-queue.tf
```diff
@@ -1,17 +1,17 @@
-module "pathfinder_events_queue" {
```
lbennett-moj: Sorry - should have removed the obsolete one on a different PR - its made the changes confusing

digitalronin: Yeah, I'm a little bit concerned that we might see odd behaviour if terraform tries to change existing resources in situ, rather than just nuking the old stuff and replacing it. It *should* be fine, but I'm not sure.
Would you mind closing this one and re-submitting as 2 separate PRs? I think that's safer for a prod. namespace.

lbennett-moj: fair enough

PR #2052, opened 2020-03-17
Update RDS module for offender-events-dev

digitalronin:


digitalronin:
Ignore this. It includes excess changes

PR #2054, opened 2020-03-17
Only sleep *between* pod deletions

digitalronin:
...instead of also sleeping after deleting the *last* pod in a namespace,
which is unnecessary


digitalronin:
Dammit - I thought I'd amended the commit to remove those. I'll fix it

digitalronin:
I have totally messed up the commit history of this PR, so I'm closing it

PR #2078, opened 2020-03-18
Require pingdom env. vars

digitalronin:


vijay-veeranki-moj:
Hi @digitalronin , plan pipeline is configured to trigger only where there is changes in below path

```
    paths:
    - namespaces/live-1.cloud-platform.service.justice.gov.uk
```

We may need to disable this "Require status checks to pass before merging" to merge this.

PR #2088, opened 2020-03-18
Add a script to find specific processes in cluster

digitalronin:
This script iterates over every container running
in the cluster, and looks for a named process
(e.g. 'python'). If it finds the process running,
it outputs the pod and container name.

The original use for this was to identify 
namespaces which might be running django 
applications, but it should work more generally.


digitalronin:
closing & reopening to kick github actions

PR #2148, opened 2020-03-23
Add make target for tools shell in the cluster

digitalronin:
Sometimes we need to be able to run terraform code from inside the same VPC
as the cluster (e.g. code which manipulates an RDS database).

To make this easier, this makefile target launches a temporary tools-image
inside the cluster,  mapping useful environment variables into it.

### DIFF
--- a/makefile
+++ b/makefile
@@ -53,4 +53,30 @@ tools-shell:
 		-v $${HOME}/.gnupg:/root/.gnupg \
 		-w /app $(TOOLS_IMAGE) bash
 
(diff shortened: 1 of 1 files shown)


digitalronin:
I upgraded a bunch of RDS modules from outside the cluster, so I think this may be unnecessary. I need to get some clarification on what connectivity we have.

digitalronin:
Actually, I think this might be useful.

Switching an RDS from random_string.password to random_password talks to the AWS API, not to the database itself. But things like postgres_role *will* need to talk to the database, which is when we'll need to run the code from within the cluster (I think)

PR #2182, opened 2020-03-26
PAT-708 - add Pathfinder IAM user with rds to s3 export permissions.

garethmdavies:
The purpose of this IAM user is to allow a future pod running in the pathfinder namespace to create an RDS to S3 snapshot export. In dev the export will be to an s3 pathfinder reporting bucket configured in the dev namespace. In production the bucket will exist in the MOJ Analytical environment. The MOJ AP will grant permission to this IAM user to put snapshots into their S3 staging bucket.

We decided not to use an assumed role like the prisoner money team do because in theory any other namespace could assume that role and export data from the DB. A specific pathfinder IAM user will contain who can perform the export task and where to.

If the reviewer could check the syntax of creating the iam user and exposing the creds to kubernetes that would be appreciated.

See: https://dsdmoj.atlassian.net/browse/PAT-708 for further details.

### DIFF
--- a/namespaces/live-1.cloud-platform.service.justice.gov.uk/pathfinder-dev/resources/rds.tf
+++ b/namespaces/live-1.cloud-platform.service.justice.gov.uk/pathfinder-dev/resources/rds.tf
@@ -1,7 +1,15 @@
+data "aws_caller_identity" "current" {}
+
+data "aws_region" "current" {}
+
 variable "cluster_name" {
 }
 
 variable "cluster_state_bucket" {
+}
+
+resource "random_id" "id" {
+  byte_length = 8
 }
 
 module "dps_rds" {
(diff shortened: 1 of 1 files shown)


namespaces/live-1.cloud-platform.service.justice.gov.uk/pathfinder-dev/resources/rds.tf
```diff
@@ -23,21 +31,63 @@ module "dps_rds" {
...
+      "rds:StartExportTask",
+      "s3:PutObject*",
+      "s3:GetObject*",
+      "s3:ListBucket",
+      "s3:DeleteObject*",
+      "s3:GetBucketLocation"
+    ]
+
+    resources = [
+      "arn:aws:rds:${data.aws_region.current.name}:${data.aws_caller_identity.current.account_id}:db:*",
```
garethmdavies: Ideally we would restrict this just to the pathfinder DB but unable to reference the db instanceId as it's not output by the RDS module.

digitalronin: That's a good point. I've opened a ticket for us to add this as an output. I don't know when that will be done, so definitely carry on with this policy for now.

garethmdavies: Thanks David, I'll keep an eye for the change.

poornima-moj: @garethmdavies,  We have added db_identifier as an output to the latest RDS module. https://github.com/ministryofjustice/cloud-platform-terraform-rds-instance/blob/master/output.tf#L42. Pls have a look if that fix the issue here.

PR #2183, opened 2020-03-26
PAT-708 - Add snapshot arn to resources

garethmdavies:
During testing it turns out we need snapshot access as its the snapshots that are extracted to s3.

Could the reviewer check the policy is correctly setup? I'm not sure if the account id and region are being pulled through correctly. I used the rds template as an example.


namespaces/live-1.cloud-platform.service.justice.gov.uk/pathfinder-dev/resources/rds.tf
```diff
@@ -59,8 +59,8 @@ data "aws_iam_policy_document" "policy" {
     resources = [
       "arn:aws:rds:${data.aws_region.current.name}:${data.aws_caller_identity.current.account_id}:rds:*",
       "arn:aws:rds:${data.aws_region.current.name}:${data.aws_caller_identity.current.account_id}:snapshot:*",
```
garethmdavies: Not sure if the region and account id are successfully populated in the final policy which may explain some access denied issues.

PR #2192, opened 2020-03-26
PAT-708 - configure iam role required for rds create-export-task command

garethmdavies:
See for further details: https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_ExportSnapshot.html
Possibility further work if export not available in eu-west-2 as indicated.

### DIFF
--- a/namespaces/live-1.cloud-platform.service.justice.gov.uk/pathfinder-dev/resources/rds.tf
+++ b/namespaces/live-1.cloud-platform.service.justice.gov.uk/pathfinder-dev/resources/rds.tf
@@ -87,6 +87,8 @@ data "aws_iam_policy_document" "pathfinder-dev-rds-to-s3-export-policy" {
 
   statement {
     actions = [
+      "rds:CopyDBSnapshot",
+      "rds:DeleteDBSnapshot",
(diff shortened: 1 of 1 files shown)


namespaces/live-1.cloud-platform.service.justice.gov.uk/pathfinder-dev/resources/rds.tf
```diff
@@ -31,43 +31,91 @@ module "dps_rds" {
...
+      "rds:DescribeDBInstances",
+      "rds:DescribeDBSnapshots",
+      "rds:DescribeExportTasks",
+      "rds:StartExportTask",
+      "rds:CancelExportTask"
+    ]
+
+    resources = [
+      "arn:aws:rds:eu-west-2:${data.aws_caller_identity.current.account_id}:db:*",
+      "arn:aws:rds:eu-west-1:${data.aws_caller_identity.current.account_id}:snapshot:*",
```
garethmdavies: Potential that we need to temporarily move the snapshot to eu-west-1 in order to perform the export to s3 followed by deleting the snapshot. This will all be scripted and ran inside a pod.

PR #2228, opened 2020-03-31
DT-618: Allow listen on expiry events

petergphillips:


namespaces/live-1.cloud-platform.service.justice.gov.uk/token-verification-api-prod/resources/elasticache.tf
```diff
@@ -13,6 +13,8 @@ module "tva_elasticache_redis" {
   team_name              = var.team_name
   number_cache_clusters  = var.number_cache_clusters
   node_type              = "cache.t2.small"
+  engine_version         = "5.0.6"
```
petergphillips, line 16: no data in there yet, don't mind if instance gets recreated due to this change

PR #2248, opened 2020-04-03
Add Logger class & use it for all command output

digitalronin:
This class acts as a single "choke point" for all
command output, so we can locate all of our log
filtering logic there, instead of having to 
duplicate it in the Terraform class.

fixes https://github.com/ministryofjustice/cloud-platform/issues/1776


digitalronin:
I got a bunch of unrelated commits mixed up in this. Closing to replace with a clean PR.

PR #2336, opened 2020-04-15
Initial setup of namespace for Contact MOJ

digitalali-moj:
I added production so that I can start using route53 as the domain so not sure exac is changing and wasn't sure of exact process for it, also so that I can work on production-specific bits if needed like pingdom and ecr. Other name spaces added too for consistency.


digitalronin:
We have asked the team to set up their dev. namespace before creating this one.

PR #2357, opened 2020-04-20
Dynamodb table for use by scheduler

andymarke:


namespaces/live-1.cloud-platform.service.justice.gov.uk/prison-to-probation-update-dev/resources/dynamodb.tf
```diff
@@ -29,3 +29,31 @@ resource "kubernetes_secret" "message_dynamodb" {
...
+
+  team_name              = var.team_name
+  application            = var.application
+  business-unit          = var.business-unit
+  environment-name       = var.environment-name
+  infrastructure-support = var.infrastructure-support
+  is-production          = "false"
+
+  hash_key       = "_id"
+  range_key      = "_id"
```
andymarke: interestingly this doesn't need a secondary index, but given range_key is mandatory I'll add it as the primary key

PR #2358, opened 2020-04-20
Dynomdb schedule for prison to probation

andymarke:


namespaces/live-1.cloud-platform.service.justice.gov.uk/prison-to-probation-update-dev/resources/dynamodb.tf
```diff
@@ -29,3 +29,31 @@ resource "kubernetes_secret" "message_dynamodb" {
...
+
+  team_name              = var.team_name
+  application            = var.application
+  business-unit          = var.business-unit
+  environment-name       = var.environment-name
+  infrastructure-support = var.infrastructure-support
+  is-production          = "false"
+
+  hash_key  = "_id"
+  range_key = "_id"
```
andymarke, line 43: interestingly this table doesn't need a secondary index, so hoping adding primary key will satisfy this *mandatory* field

PR #2365, opened 2020-04-21
Increase Form Builder Platform request limits

brenetic:
Screenshot 2020-04-21 at 09 29 57

The requests memory limits seem to always be in the red in production.
Raise the limit to 256MB to give a little more headroom.

https://trello.com/c/2Pgew3OG/341-increase-memory-allowance-in-platform-apps


digitalronin:
@brenetic bear in mind that "defaultRequest" is the amount of resources each container *reserves* when it starts up. It doesn't actually increase the limit of what your containers are *allowed* to use - that's the "default" value on line 13.

digitalronin:
After discussing this with the team, this change isn't actually needed. Their containers are not getting close to the limit on the memory they're allowed to *use* - the grafana dashboard was misleading. So, we've agreed to close this PR.

PR #2376, opened 2020-04-22
Add offender search KIAM roles to preprod and prod namespaces

jhevans:


jhevans:
Set as draft to prevent merge while waiting on confirmation from service owners to allow this access in preprod/prod envs.

digitalronin:
@jhevans Is this still relevant?

jhevans:
@digitalronin yeah this is still needed. It's taken a while to get the all clear from the service owners but we have that now so are planning to pick this work up again soon

PR #2389, opened 2020-04-23
create common platform queues in test env

praveenmoj:


praveenmoj:
> You've got the "multiple queues cyclic dependency" problem described here:
> https://github.com/ministryofjustice/cloud-platform-terraform-sqs#usage
> 
> You need to split this into 2 PRs by commenting out all but one of the queues, so that terraform knows which queue will _create_ the IAM user, and the other queues will all _reuse_ that IAM user.
> 
> As it stands, terraform has no way of knowing which is the queue that will create the IAM user, so it can't create _any_ of the queues.

PR #2478, opened 2020-05-21
Deploy version 2.3 of the HOODAW web app.

digitalronin:


digitalronin:
☝️ the check failure is expected, because the secrets.yaml file is git-crypted, and the "malformed yaml" check doesn't decrypt the repo before checking.

PR #2489, opened 2020-06-01
Add script to find deprecated kubernetes objects

digitalronin:
This script will output a CSV report listing all objects which
use API versions deprecated in kubernetes 1.16

It also lists *all* tiller pods (not just those which are too old)

fixes https://github.com/ministryofjustice/cloud-platform/issues/1942

### DIFF
--- a/bin/find-deprecated-objects.rb
+++ b/bin/find-deprecated-objects.rb
@@ -106,7 +106,7 @@ def object_csv(hash, namespaces)
     hash.dig("metadata", "name"),
     namespace,
     team,
-    repo,
+    repo
   ].join(", ")
 end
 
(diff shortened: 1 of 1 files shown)


bin/find-deprecated-objects.rb
```diff
@@ -0,0 +1,129 @@
...
+    end
+  end
+
+  rtn
+end
+
+def deprecated_api_objects
+  # NB: double backslashes, compared to running the command from a terminal
+  cmd = %(
+    kubectl get NetworkPolicy,PodSecurityPolicy,DaemonSet,Deployment,ReplicaSet --all-namespaces
```
poornima-moj, line 74: @digitalronin , can we also include "Ingress" here.

digitalronin: Good idea, @poornima-moj 

I've made that change.

PR #2490, opened 2020-06-02
Add laa-cla-backend-staging and laa-cla-backend-production rds instances

said-moj:
Add rds instance to the following namespaces:
- laa-cla-backend-staging
- laa-cla-backend-production

### DIFF
--- a/namespaces/live-1.cloud-platform.service.justice.gov.uk/laa-cla-backend-staging/resources/rds.tf
+++ b/namespaces/live-1.cloud-platform.service.justice.gov.uk/laa-cla-backend-staging/resources/rds.tf
@@ -1,0 +1,60 @@
+/*
+ * Make sure that you use the latest version of the module by changing the
+ * `ref=` value in the `source` attribute to the latest version listed on the
+ * releases page of this repository.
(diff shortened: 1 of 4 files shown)


namespaces/live-1.cloud-platform.service.justice.gov.uk/laa-cla-backend-staging/resources/rds.tf
```diff
@@ -0,0 +1,60 @@
...
+
+  db_name = "cla_backend"
+  # change the postgres version as you see fit.
+  db_engine_version      = "9.6"
+  environment-name       = var.environment-name
+  infrastructure-support = var.infrastructure-support
+
+  # rds_family should be one of: postgres9.4, postgres9.5, postgres9.6, postgres10, postgres11
+  # Pick the one that defines the postgres version the best
+  rds_family = "postgres96"
```
digitalronin, line 31: This needs to be `postgres9.6` with a dot between 9 and 6

the plan pipeline check doesn't pick this up, but it will fail when terraform actually makes calls to the AWS API.

PR #2501, opened 2020-06-04
Skeleton namespace for Prisoner Content Hub

spikeheap:
This PR creates a namespace for the Prisoner Content Hub, which we're going to migrate from our Azure infrastructure to Cloud Platform.

- Creates namespace with standard limits, NetworkPolicy, etc. (cloned from DPS).
- Adds Terraform skeleton & variables ready for resources to follow.

We have assumed that the `rds_family` attribute for MariaDB would be `mariadb10.4`, which matches the pattern for MySQL, but not Postgres. We couldn't find a list of the family types, but assume it's required.


spikeheap:
Thanks @digitalronin. I've removed the RDS portion of this PR. 

Given that the MoJ module only supports Postgres, would you like a PR to that project's docs to make that clear (or is it something you're actively looking to widen to MySQL/MariaDB)?

digitalronin:
> Thanks @digitalronin. I've removed the RDS portion of this PR.
> 
> Given that the MoJ module only supports Postgres, would you like a PR to that project's docs to make that clear (or is it something you're actively looking to widen to MySQL/MariaDB)?

A doc PR would be very welcome.

We did look at extending to mysql before, to accomodate Wordpress sites. I can't remember why we ended up not going ahead with that.

Given how rarely we use anything except Postgres in production, it's probably simpler to just leave that module as it is, but you're right - we should make it clear in the docs (and rename it), if it's just for Postgres.

PR #2512, opened 2020-06-08
Use 'pluto' in script to find deprecated objects

digitalronin:
This PR updates the `bin/find-deprecated-objects.rb` script to use
[pluto](https://github.com/FairwindsOps/pluto) to report on objects created via
Helm 2 & 3.

### DIFF
--- a/bin/find-deprecated-objects.rb
+++ b/bin/find-deprecated-objects.rb
@@ -92,7 +92,8 @@ end
 # Given a hash representing a kubernetes object, return a csv of the fields we
 # care about wrt. the API deprecation
 def object_csv(hash, namespaces)
-  namespace, team, repo = namespace_team_repo(hash, namespaces)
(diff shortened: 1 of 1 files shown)


bin/find-deprecated-objects.rb
```diff
@@ -24,17 +26,68 @@ def main
...
+  end
+end
+
+def helm_object_csv(object, namespaces)
+  team, repo = namespace_team_repo(object["namespace"], namespaces)
+
+  # 'name' is often something like "check-financial-eligibility/check-financial-eligibility"
+  # so if it matches that pattern, simplify it.
+  name = object.fetch("name")
+  /(.*)\/\1/.match(name) && name = $1
```
poornima-moj, line 71: @digitalronin , For now, it is good and shows the helm release name. But for later, to either display the whole name so we know the helm release name as well as resource name.

PR #2529, opened 2020-06-09
Update 02-limitrange.yaml

nickmcmahon01:


nickmcmahon01:
Taking the values from https://github.com/ministryofjustice/cloud-platform-environments/blob/master/namespaces/live-1.cloud-platform.service.justice.gov.uk/sentence-planning-development/02-limitrange.yaml where we had no problem running springboot apps

PR #2538, opened 2020-06-10
Adding cors rule for static files bucket for production

said-moj:


said-moj:
> This is an S3 bucket that we're already added to the "allow public" list, right?
Yep

PR #2564, opened 2020-06-15
added route53 zone for hmpps-submit-information-report-dev

whitfield-mj:


namespaces/live-1.cloud-platform.service.justice.gov.uk/hmpps-submit-information-report-dev/resources/variables.tf
```diff
@@ -1,3 +1,7 @@
+variable "domain" {
+  default = "hmpps-submit-information-report-dev.service.justice.gov.uk"
```
mtrbls, line 2: I'd suggest creating `hmpps-submit-information-report.service.justice.gov.uk` instead.
You'll be able to create `dev.hmpps-submit-information-report.service.justice.gov.uk` with an ingress resource later.

whitfield-mj: We've been going round and round with this idea internally. It has been requested that dev is appended as shown. I will raise it again

PR #2583, opened 2020-06-17
Ruby script to list and update files which has force_ssl and apply_me…

poornima-moj:
…thod strings. 

To update RDS module to 5.4

Related to: https://github.com/ministryofjustice/cloud-platform/issues/1563

### DIFF
--- a/bin/rds-force_ssl-list-update.rb
+++ b/bin/rds-force_ssl-list-update.rb
@@ -1,10 +1,10 @@
 #!/usr/bin/env ruby
 
-# This script takes a action and option
-# act: list, update
+# This script takes two inputs, `action` and `option`
+# action: list, replace
 # option: force_ssl_true, force_ssl_false, apply_pending_reboot
 # 
-# This script can do 2 things based on the action
+# This script can do 2 things based on the `action`
 # - list all terraform files inside all directories inside NAMESPACE_DIR
 #    which has force_ssl = "true", force_ssl = "false" or apply_method = "pending-reboot"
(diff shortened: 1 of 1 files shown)


bin/rds-force_ssl-list-update.rb
```diff
@@ -0,0 +1,97 @@
...
+    stdout.split("\n")
+  end
+end
+
+def tfdir(namespace)
+  "#{NAMESPACE_DIR}/#{namespace}/resources"
+end
+
+############################################################
+# a = force_ssl = true, b = force_ssl = "true", c = force_ssl = "false"
```
digitalronin, line 95: I don't understand what this comment line is trying to tell me

bin/rds-force_ssl-list-update.rb
```diff
@@ -0,0 +1,97 @@
+#!/usr/bin/env ruby
+
+# This script takes a action and option
+# act: list, update
```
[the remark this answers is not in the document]
poornima-moj: Updated the comment

bin/rds-force_ssl-list-update.rb
poornima-moj: Its a copy paste comment. Not really needed so removed

poornima-moj:
> This looks good, Poornima. I've made a couple of small suggestions below.
> 
> Also, you use both "action" and "act" to refer to the same thing, in the code and in the comments. I'd recommend picking one and using it consistently - it will make the script easier to understand.

@digitalronin , thanks for your comments. I have used "action" consistently and update the code.

PR #2591, opened 2020-06-18
PAT-998: Add DNS record for manage-soc-cases-dev

timharrison-moj:
As this service is only present in dev (until it undergoes pen test & DPIA process), the dev instance requires a DNS entry for ingress.


timharrison-moj:
Thanks David - yes, we've been around the houses on the naming, and this is the one settled upon. Before this gets into a live environment I suspect we will agree upon a single domain for services which are neither prison nor probation specific eg. .hmpps.service.justice.gov.uk, and have sub-domains for services and environments, but its not been agreed in the TA community yet.  On the particular name 'manage-soc-cases' - this was a compromise made with the backing of the product owner after full and frank discussions with the content designers, who wanted a name which would wrap across the screen.

PR #2614, opened 2020-06-23
Pic 489 cpmg secrets

chris-faulkner-mojdt:
Need some secrets for auth to JMS and the admin console of a running wildfly instance. I haven't done this before - do I need anything to refer to "wildfly-creds.tf"?

### DIFF
--- a/namespaces/live-1.cloud-platform.service.justice.gov.uk/crime-portal-mirror-gateway-dev/resources/wildfly-creds.tf
+++ b/namespaces/live-1.cloud-platform.service.justice.gov.uk/crime-portal-mirror-gateway-dev/resources/wildfly-creds.tf
@@ -1,0 +1,23 @@
+resource "random_password" "adminuser_password" {
+  length  = 16
+  special = true
+}
+
+resource "random_password" "jmsuser_password" {
+  length  = 16
+  special = true
+}
+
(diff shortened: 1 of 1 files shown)


digitalronin:
Hi Chris

Just so I understand this - you're going to populate some random credentials in your namespace, and then manually set up accounts on the target systems which use those credentials. Is that right?

If that's correct, then you can totally do this, although you will need to manually rotate the credentials in the target system if you ever update them in your namespace.

Another option would be to manage the secrets in your repo using git-crypt - i.e. set up the accounts in the target system, and define a "secret.yaml" file in your own repository which creates a kubernetes secret in your namespace containing the credentials. This is effectively the same as what you're doing here - it's just a difference in where/how you manage those secrets.

chris-faulkner-mojdt:
> Hi Chris
> 
> Just so I understand this - you're going to populate some random credentials in your namespace, and then manually set up accounts on the target systems which use those credentials. Is that right?
> 
> If that's correct, then you can totally do this, although you will need to manually rotate the credentials in the target system if you ever update them in your namespace.
> 
> Another option would be to manage the secrets in your repo using git-crypt - i.e. set up the accounts in the target system, and define a "secret.yaml" file in your own repository which creates a kubernetes secret in your namespace containing the credentials. This is effectively the same as what you're doing here - it's just a difference in where/how you manage those secrets.

PR #2633, opened 2020-06-24
Define missing settings for read replica.

mafonso:


poornima-krishnasamy:
Reopening to check the plan pipeline

PR #2657, opened 2020-06-29
Revert change to add a slack channel to namespaces

digitalronin:
This change breaks PRs against existing namespaces, so I'm reverting it
until I can fix that.


digitalronin:
Never mind. #2658 should fix this.

PR #2683, opened 2020-07-02
Changes to allow cloud-portal-mirror-gateway to be deployed to court-probation-dev

jhevans:


chris-faulkner-mojdt:
I think that this is a tactical solution whilst our inter-namespace communication issues are sorted out. In the context of that, it makes sense I think for now. (?)

> Have you considered "crime-portal-mirror-gateway.service.justice.gov.uk" in the context of our guidance for [naming things](https://ministryofjustice.github.io/technical-guidance/standards/naming-things/#naming-things)?
> 
> If you have, then fine, but as an uninformed citizen, that domain name isn't very clear to me.

jhevans:
> Have you considered "crime-portal-mirror-gateway.service.justice.gov.uk" in the context of our guidance for [naming things](https://ministryofjustice.github.io/technical-guidance/standards/naming-things/#naming-things)?
> 
> If you have, then fine, but as an uninformed citizen, that domain name isn't very clear to me.

@digitalronin I tend to agree but the name has already been agreed with a third party and we're short on time to get it ready for testing. It is only ever due to be used by this one 3rd party and they originally named it so it makes sense to them fwiw. It also has a relatively short planned service life -  we're hoping to move to a more flexible API when one becomes available. I'll admit that is a little speculative though 😏 maybe next year... maybe not!
 
If we can get this through this first round of testing we'd be happy to discuss it further, we know we'll have some further tweaks to do to it anyway.

PR #2690, opened 2020-07-02
✨ Create new service account for helm3 in court-probation-dev namespace

jhevans:


namespaces/live-1.cloud-platform.service.justice.gov.uk/court-probation-dev/07-serviceaccount-circlecihelm3.yaml
```diff
@@ -0,0 +1,19 @@
...
+metadata:
+  name: circlecihelm3
+  namespace: court-probation-dev
+subjects:
+- kind: ServiceAccount
+  name: circlecihelm3
+  namespace: court-probation-dev
+roleRef:
+  kind: Role
+  name: admin
```
digitalronin, line 18: Shouldn't this be "circleci", to reference this role:
https://github.com/ministryofjustice/cloud-platform-environments/blob/59218de3757ed09d0393722d263297f330a9f568/namespaces/live-1.cloud-platform.service.justice.gov.uk/court-probation-dev/05-serviceaccount-circleci.yaml#L10

jhevans: No this is deliberate, we're using helm3 which apparently needs different perms to helm2 which we've been using up to now in this namespace so for the existing user with that role we get the following error:
```Error: query: failed to query with labels: secrets is forbidden: User "system:serviceaccount:*******************:circleci" cannot list resource "secrets" in API group "" in the namespace "*******************"
```

I appreciate the admin role is probably overly permissive but I've not been able to find a full list of the permissions that helm3 will need to build an exact role profile.

jhevans: The only other namespace I know of that this works with has admin as the role hence choosing that

jhevans: I may have misunderstood the issue there - there was an error, should have been ClusterRole not Role

jhevans:
Sorry... have just found another discrepancy so have had to update

PR #2706, opened 2020-07-07
Add new namespace laa-court-data-ui-uat

jsugarman:
Add new UAT namespace laa-court-data-ui-uat

Inline with API and other components of
common platform interfaces for LAA.


jsugarman:
Do i need to break this up a bit or can it be done in one-shot?

PR #2708, opened 2020-07-07
🔧  Update NetworkPolicy for crime-portal-mirror-gateway-preprod and al…

jhevans:
…ign crime-portal-mirror-gateway-prod with lower environments

### DIFF
--- a/namespaces/live-1.cloud-platform.service.justice.gov.uk/crime-portal-mirror-gateway-dev/resources/ecr.tf
+++ b/namespaces/live-1.cloud-platform.service.justice.gov.uk/crime-portal-mirror-gateway-dev/resources/ecr.tf
@@ -46,10 +46,10 @@ resource "kubernetes_secret" "pict_cpmg_wildfly_ecr_credentials" {
   }
 }
 
-resource "kubernetes_secret" "pict_cpmg_database_ecr_credentials_court_probation_dev" {
+resource "kubernetes_secret" "pict_cpmg_database_ecr_credentials_preprod" {
   metadata {
     name      = "pict-cpmg-ecr-database-credentials-output"
-    namespace = "court-probation-dev"
+    namespace = "crime-portal-mirror-gateway-preprod"
   }
 
   data = {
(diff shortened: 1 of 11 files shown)


namespaces/live-1.cloud-platform.service.justice.gov.uk/crime-portal-mirror-gateway-dev/resources/ecr.tf
```diff
@@ -46,10 +46,10 @@ resource "kubernetes_secret" "pict_cpmg_wildfly_ecr_credentials" {
   }
 }
 
-resource "kubernetes_secret" "pict_cpmg_database_ecr_credentials_court_probation_dev" {
+resource "kubernetes_secret" "pict_cpmg_database_ecr_credentials_preprod" {
```
chris-faulkner-mojdt, line 49: This is in the crime-portal-mirror-gateway-dev namespace looking at the folder ?

jhevans: Yeah this is a bit of an odd one. The ECR strictly isn't really related to a specific namespace, its just some generic AWS stuff which we need access to in all of our envs. It has to be created somewhere though so it's ended up in dev. We need those creds in all 3 namespaces though so I've had to put it here. 

@digitalronin I don't suppose you can advise of a preferred way of declaring shared resources like this between namespaces?

jhevans: > We're still having one ECR with one set of creds ?
I don't see what we gain from duplicating it to be honest, the artefact will be identical so if we did it would be purely to make it fit with our 3 namespaces. We'd then have to push the image to 3 different ECRs with 3 different sets of credentials from CircleCI adding more complexity and maintenance overhead. Also we'd be paying for 3 instances when we only need one.

Maybe it would be worth documenting what's going on at least though. I could add an explanatory comment to the ecr terraform?

digitalronin: The question of sharing across namespaces has started to come up more frequently, but unfortunately we don't have a very good answer yet.
I think it really depends what you're trying to do by "declaring" the shared resource. If it's just to share the knowledge of it, you could do something as simple as putting a README file in the other environment folders for namespaces which depend on the shared resource.
If you need to share secrets you can add some terraform code in the folder where the secret is created, and tell that code to push the secret into the other namespaces where you want it. Because the terraform code runs in our pipeline, with root privileges, that will work. It's not ideal for lots of reasons, but so far these cases are still relatively rare, even though they are on the increase.

jhevans: Thanks, I've added an explanatory comment to document it as best I can and update so it's at least obvious what's going on.

PR #2751, opened 2020-07-10
remove redundant role since no longer bridging to Delius AWS

andymarke:


namespaces/live-1.cloud-platform.service.justice.gov.uk/offender-search-dev/00-namespace.yaml
```diff
@@ -10,4 +10,3 @@ metadata:
     cloud-platform.justice.gov.uk/application: "offender-search"
     cloud-platform.justice.gov.uk/owner: "Digital Prison Services: dps-hmpps@digital.justice.gov.uk"
     cloud-platform.justice.gov.uk/source-code: "https://github.com/ministryofjustice/offender-search.git"
-    iam.amazonaws.com/permitted: "offender-search-elastic-search-delius-dev"
```
andymarke, line 13: No longer needed since now using CP Elastic Search not Delius Elastic Search

PR #2755, opened 2020-07-10
Template files referenced when creating namespace using cloud-platform CLI

poornima-krishnasamy:
These template files are used by the [cloud-platform-cli ](https://github.com/ministryofjustice/cloud-platform-moj-cp) to create namespace for end users of cloud platform

Related to: https://github.com/ministryofjustice/cloud-platform/issues/1961


poornima-krishnasamy:
> I think we should decide, before merging this, if we're going to go ahead with the golang cli tool. If we are, we should replace the templates in "namespace-resources" and "gitops-namespace-resources", rather than adding a third set of templates to maintain.

@digitalronin , shall we merge this folder so that the cli code and the tools image can use the main branch. Eventually, when ready we can change the makefile to use cli instead of the ruby script and then remove the other folder.

PR #2774, opened 2020-07-17
Update authorized-keys-provider s3 module to 4.3

digitalronin:
This namespace seems to have been missed in #2772


digitalronin:
There might be a reason this namespace was left out of #2772 
Check this [slack thread](https://mojdt.slack.com/archives/C514ETYJX/p1594976333292400) before merging.

digitalronin:
This namespace was left out of #2772 so that the plan pipeline would not reject that PR because of the encrypted yaml file in this namespace folder (which causes the "malformed yaml" check to fail). 
It's safe to merge this now.

PR #2797, opened 2020-07-21
Book secure move dev fixes

mafonso:
Fixing  S3 permissions.


namespaces/live-1.cloud-platform.service.justice.gov.uk/hmpps-book-secure-move-api-dev/resources/rds.tf
```diff
@@ -70,7 +70,6 @@ module "rds-read-replica" {
   ]
 }
 
-
```
mafonso, line 73: Terraform fmt

PR #2868, opened 2020-07-30
Update script to Create branch and raise PR only if new keys are present

poornima-krishnasamy:


bin/update-authorized-keys.rb
```diff
@@ -48,41 +55,55 @@ def build_authorized_keys
...
+    # Get the main branch sha
+    json = get_query(
+      token: ENV.fetch("GITHUB_TOKEN"),
+      api_url: GITHUB_REST_API_URL + "/git/ref/heads/main"
+    )
+  
+    sha = JSON.parse(json).dig("object", "sha")
+  
+    # Create a reference which is basically creating a branch
+    json = post_query(
```
digitalronin: Are you doing anything with the "json = " value from the post and put queries? If not, it's best not to assign the return values at all. i.e. 

```
post_query(...)
```

Instead of 

```
json = post_query(...)
```

poornima-krishnasamy: Good point @digitalronin. I have removed capturing return values.

PR #2877, opened 2020-07-31
DT-1055: 🔨 Initial stab at adding ingress controller

petergphillips:


petergphillips:
Found out that I can edit the ingress definition to add our custom header to nginx instead, since it is nginx and not the NLB that is terminating the SSL.

PR #2940, opened 2020-08-18
Update rds.tf in dstest namespace

digitalronin:
depends on
https://github.com/ministryofjustice/cloud-platform-terraform-rds-instance/pull/71


digitalronin:
Plan says this change makes no difference, which is what I wanted to confirm, so closing this.

PR #2942, opened 2020-08-19
⬆️ Upgrade terraform version to ensure latest compatible version is u…

jhevans:
…sed in court-probation-dev

### DIFF
--- a/namespaces/live-1.cloud-platform.service.justice.gov.uk/court-probation-dev/resources/versions.tf
+++ b/namespaces/live-1.cloud-platform.service.justice.gov.uk/court-probation-dev/resources/versions.tf
@@ -1,4 +1,4 @@
 
 terraform {
-  required_version = "0.12.17"
+  required_version = ">= 0.12"
 }


namespaces/live-1.cloud-platform.service.justice.gov.uk/court-probation-dev/resources/versions.tf
```diff
@@ -1,4 +1,4 @@
 
 terraform {
-  required_version = "0.12.17"
+  required_version = ">= 0.12"
```
vijay-veeranki, line 3: @jhevans , we are using terraform version "0.12.17", across the cluster. 

We want to maintain same version across all the namespaces as the pipeline is using "0.12.17" to apply terraform.

vijay-veeranki (changes requested):
Any reason why you want to change to ">= 0.12"?

namespaces/live-1.cloud-platform.service.justice.gov.uk/court-probation-dev/resources/versions.tf
jhevans: We received an advisory to upgrade from the pen testers so I figured if I applied this we would get any future patch upgrades. If that's the case you're right there's no point in applying these changes. Will close

PR #2945, opened 2020-08-19
⬆️ Upgrade terraform version to ensure latest compatible version is u…

jhevans:
…sed in court-probation-prod


vijay-veeranki (changes requested):
@jhevans , Any reason why you want to change to ">= 0.12" now?.
we are using terraform version "0.12.17", across the cluster.
We want to maintain same version across all the namespaces as the pipeline is using "0.12.17" to apply terraform.

We have a ticket to upgrade to terraform 1.13, then we will upgrade every namespace to use that version.

PR #2961, opened 2020-08-24
Add a script to count ingresses by ingress class

digitalronin:
fixes https://github.com/ministryofjustice/cloud-platform/issues/2245


jasonBirchall (approved):
Part of me thinks this being a 22 line script is a bit excessive when the same could be achieved by a one line command. I'll approve anyway as the code looks good. :+1:

digitalronin:
> Part of me thinks this being a 22 line script is a bit excessive when the same could be achieved by a one line command. I'll approve anyway as the code looks good. 👍

It could absolutely be done in a one-liner, but I always prefer clarity over conciseness, for reasons I'm happy to bore people about at any time ;)

PR #2965, opened 2020-08-24
Add a dedicated ingress to mod-sec namespaces

digitalronin:
These are the namespaces which use mod-security.
So, these are the ones we need to move to
dedicated ingress controllers as a matter of
urgency

This PR will create a separate AWS load-balancer
and ingress controller for each namespace, so that
the teams can add the ingress class annotation to
their ingress definitions to start routing traffic
through the new ingresses.

As soon as all of these namespaces have migrated
their traffic, we will disable mod-security on the
default ingress, which should reduce the problems
we are seeing where the ingress pods are running
out of shared memory.

Related to https://github.com/ministryofjustice/cloud-platform/issues/2224


digitalronin:
Added "do not merge" label because we want to merge this in the morning. That way we'll be able to respond quickly if, e.g. we need to increase any namespace limits or add more worker nodes.

digitalronin:
closing in favour of #2966 Let's do non-prod first, just in case.

PR #2982, opened 2020-08-26
🚀 Create new wildcard certificate in court-probation-dev namespace

jhevans:


digitalronin (changes requested):
This isn't going to work. We've hit a rate limit at LetsEncrypt for certificates with the domain name `*.apps.live-1.cloud-platform.service.justice.gov.uk`

Is there any way to change the domain to `something-specific.live-1.cloud-platform.service.justice.gov.uk` ?

jhevans:
> This isn't going to work. We've hit a rate limit at LetsEncrypt for certificates with the domain name `*.apps.live-1.cloud-platform.service.justice.gov.uk`
> 
> Is there any way to change the domain to `something-specific.live-1.cloud-platform.service.justice.gov.uk` ? (NB: There's no `apps` part in there)

Yes I think that should be possible but I'll need to prime the team first so nobody is wondering why it's disappeared 😄 
Should we avoid using wildcards as well? i.e. do I need to create certificates for each service in the namespace?

jhevans:
@digitalronin in retrospect, though we can make the changes it's going to have knock on effects and will be more time consuming than we'd originally anticipated. I'm going to close this PR for now whilst we work out how we're going to go about it.

PR #3004, opened 2020-08-27
Add c100-application-metabase ingress controller

zheileman:
Same as with previous PR #3003 but in this namespace we didn't have yet a `resources` dir so I've added the necessary files to it with the basic config and variables that we need.


jasonBirchall (approved):
Looks good to me. :+1:

PR #3007, opened 2020-08-28
PAT 1230 - ingress controllers for pathfinder namespaces + small changes to manage-soc-cases for consistency

timharrison-moj:
* Adds ingress controllers to pathfinder-dev, -preprod and -prod namespaces.
* Small changes to manage-soc-cases - to make it consistent with others
* Removed some unused variables
* Set is_production = var.is-production in all ingress definitions

### DIFF
--- a/namespaces/live-1.cloud-platform.service.justice.gov.uk/manage-soc-cases-prod/resources/ingress.tf
+++ b/namespaces/live-1.cloud-platform.service.justice.gov.uk/manage-soc-cases-prod/resources/ingress.tf
@@ -1,6 +1,6 @@
 module "ingress_controller" {
   source = "github.com/ministryofjustice/cloud-platform-terraform-teams-ingress-controller?ref=0.0.7"
 
-  namespace = "manage-soc-cases-prod"
+  namespace     = "manage-soc-cases-prod"
   is_production = var.is_production
 }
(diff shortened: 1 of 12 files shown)


namespaces/live-1.cloud-platform.service.justice.gov.uk/manage-soc-cases-prod/resources/ingress.tf
```diff
@@ -2,5 +2,5 @@ module "ingress_controller" {
   source = "github.com/ministryofjustice/cloud-platform-terraform-teams-ingress-controller?ref=0.0.7"
 
   namespace     = "manage-soc-cases-prod"
-  is_production = var.is_production
+  is_production = var.is-production
```
timharrison-moj, line 5: Re-use existing value rather than adding a new variable.

namespaces/live-1.cloud-platform.service.justice.gov.uk/manage-soc-cases-prod/resources/variables.tf
```diff
@@ -30,25 +30,6 @@ variable "is-production" {
...
-
-variable "infrastructure_support" {
-  description = "The team responsible for managing the infrastructure. Should be of the form team-email."
-  default     = "dps-hmpps@digital.justice.gov.uk"
-}
-
-variable "is_production" {
-  default = "true"
-}
-
```
timharrison-moj, line 51: Don't think we need these for ingress - not used in that module.

digitalronin (commented):
I would recommend leaving the `is_production` var in place.

We recently added this ADR, because our naming was inconsitent:
https://github.com/ministryofjustice/cloud-platform/blob/main/architecture-decision-record/017-Variable-Naming.md

Any future modules we write/modify will use snake-case versions of variable names, and we are going to go through all namespaces and change the terraform code to *only* have snake-case variants. So, I think it's better to start using those ASAP.

However, if you feel strongly about it, I'll approve this PR as is.

digitalronin (approved):

PR #3016, opened 2020-09-01
Use var.namespace in ingress.tf

digitalronin:
We don't need to do any template interpolation here - we can just use the existing terraform variable to supply the namespace at plan/apply time.


jasonBirchall (approved):
This seems sensible to me.

PR #3023, opened 2020-09-02
Apply sonar scanner

imran-awan1:


digitalronin (changes requested):
I think this ruby code can be optimised a bit. Let's have a call and I can show you what I mean.

digitalronin (approved):

PR #3034, opened 2020-09-03
DT-1086: 🔧 Add dedicated ingress controllers

petergphillips:


poornima-krishnasamy (commented):
@petergphillips , can u add  main.tf and version.tf to this folder.

poornima-krishnasamy (approved):

PR #3044, opened 2020-09-04
Update ingress modules to v0.1.0

digitalronin:
```
ag -l "github.com/ministryofjustice/cloud-platform-terraform-teams-ingress-controller.ref=0.0.9" | xargs sed -i '' 's/ref=0.0.9/ref=0.1.0/'
```


digitalronin:
This PR would restart all the affected ingresses, which could cause some downtime. We need to think about how to manage this.

digitalronin:
Closing this, because we need to investigate why non-prod namespaces saw 4 minutes of downtime when they were updated (assuming that's the cause of the downtime).
https://mojdt.slack.com/archives/C57UPMZLY/p1599228156358200

PR #3045, opened 2020-09-04
Update c100 prod ingress to module version 0.1.0

digitalronin:
This change should not incur any application downtime.


vijay-veeranki (approved):

digitalronin:
Closing this, because we need to investigate why non-prod namespaces saw 4 minutes of downtime when they were updated (assuming that's the cause of the downtime).
https://mojdt.slack.com/archives/C57UPMZLY/p1599228156358200

digitalronin:
For future reference: This particular upgrade uses a different Helm chart for the ingress controller. This results in the whole ingress controller being recreated when this PR is applied, including a new AWS load-balancer and a DNS change to point to it. Hence the downtime.

PR #3055, opened 2020-09-07
NN-2906: Add dedicated ingress controllers

Nimphal:
This covers:
1. Digital Prison Services namespace
2. Keyworker API namespace
3. Manage Key Workers namespace
4. Whereabouts API namespace


digitalronin (changes requested):
The ingress controller module is now on version 0.1.0  Please use that version, for this PR.

Nimphal:
> The ingress controller module is now on version 0.1.0 Please use that version, for this PR.

Updated :)

digitalronin (approved):

PR #3136, opened 2020-09-15
Updated the policy name for a SQS create link sp status

atharmajeed:


digitalronin:
@atharmajeed Please could you update this PR to resolve the merge conflicts?

atharmajeed:
> @atharmajeed Please could you update this PR to resolve the merge conflicts?

We can close this PR as it does not require. Initially, I thought the queue name length was causing the error, but it turns out to be not the case. If you agree with this, then we don't need to merge this PR.

PR #3168, opened 2020-09-21
Add laa-cla-backend-training namespace

exonian:
Based on the staging namespace, with namespace values updated and
is-production set to true as, while this service is not for the public,
it is used by call centre operators who have an expectation of up time.


digitalronin:
@exonian the plan pipeline shows some problems with this PR. Please could you take a look?

mtrbls (approved):

PR #3210, opened 2020-09-25
Delete file in incorrect location

sablumiah:
Delete file in incorrect location


jasonBirchall:
Why do you think it's in an incorrect location?

PR #3420, opened 2020-10-02
🔧 PPD-47: Adding separate IAM user for uploading to S3

brightonsbox:
The intention here is to demonstrate that we can separate the permissions of our application (which should only be able to get and delete objects from S3) from the permissions of another IAM user which can only put objects to the bucket.

### DIFF
--- a/namespaces/live-1.cloud-platform.service.justice.gov.uk/hmpps-pin-phone-monitor-dev/resources/iam.tf
+++ b/namespaces/live-1.cloud-platform.service.justice.gov.uk/hmpps-pin-phone-monitor-dev/resources/iam.tf
@@ -1,0 +1,65 @@
+resource "random_id" "upload" {
+  byte_length = 8
+}
+
+resource "aws_iam_user" "upload_user" {
+  name = "hmpps-pin-phone-monitor-upload-user-${random_id.upload.hex}"
+  path = "/system/hmpps-pin-phone-monitor-upload-users/"
+}
+
+resource "aws_iam_access_key" "upload_user" {
+  user = aws_iam_user.upload_user.name
+}
+
+data "aws_iam_policy_document" "upload_policy" {
+  statement {
+    actions = ["s3:PutObject"]
+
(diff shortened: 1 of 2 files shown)


namespaces/live-1.cloud-platform.service.justice.gov.uk/hmpps-pin-phone-monitor-dev/resources/iam.tf
```diff
@@ -0,0 +1,65 @@
...
+
+    resources = [
+      "${module.hmpps_pin_phone_monitor_document_s3_bucket.bucket_arn}",
+      "${module.hmpps_pin_phone_monitor_document_s3_bucket.bucket_arn}/*"
+    ]
+
+    condition {
+      test     = "Bool"
+      variable = "aws:SecureTransport"
+      values   = ["true"]
```
garethmdavies, line 37: think this should be false

brightonsbox: Good spot, fixed 👍

namespaces/live-1.cloud-platform.service.justice.gov.uk/hmpps-pin-phone-monitor-dev/resources/iam.tf
```diff
@@ -0,0 +1,65 @@
...
+    effect = "Deny"
+
+    actions = ["s3:*"]
+
+    resources = [
+      "${module.hmpps_pin_phone_monitor_document_s3_bucket.bucket_arn}",
+      "${module.hmpps_pin_phone_monitor_document_s3_bucket.bucket_arn}/*"
+    ]
+
+    condition {
```
brightonsbox, line 34: Deny any attempts to upload over unsecure (non-HTTPS) connection

jasonBirchall (approved):

PR #3421, opened 2020-10-02
one route53 zone per domain in the list

mtrbls:
Leaving tribunals.gov.uk there for now.

### DIFF
--- a/namespaces/live-1.cloud-platform.service.justice.gov.uk/maintenance-pages/resources/variables.tf
+++ b/namespaces/live-1.cloud-platform.service.justice.gov.uk/maintenance-pages/resources/variables.tf
@@ -71,3 +71,32 @@ variable "slack-channel" {
   default     = "cloud-platform"
 }
 
+
+variable "domains" {
+  description = "List of domains to be put in maintenance"
+  type = any
+  default = [
+    "adminappeals.reports.tribunals.gov.uk",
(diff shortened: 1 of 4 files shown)


namespaces/live-1.cloud-platform.service.justice.gov.uk/maintenance-pages/resources/variables.tf
```diff
@@ -71,3 +71,32 @@ variable "slack-channel" {
...
+    "claimsmanagement.decisions.tribunals.gov.uk",
+    "consumercreditappeals.decisions.tribunals.gov.uk",
+    "employmentappeals.decisions.tribunals.gov.uk",
+    "estateagentappeals.decisions.tribunals.gov.uk",
+    "financeandtax.decisions.tribunals.gov.uk",
+    "immigrationservices.decisions.tribunals.gov.uk",
+    "informationrights.decisions.tribunals.gov.uk",
+    "landregistrationdivision.decisions.tribunals.gov.uk",
+    "landschamber.decisions.tribunals.gov.uk",
+    "maintenance-pages-demo.apps.live-1.cloud-platform.service.justice.gov.uk",
```
digitalronin, line 93: This is a duplicate line, and we don't need this anyway, do we? Doesn't the existing live-1 wildcard certificate handle this already?

digitalronin (changes requested):

PR #3425, opened 2020-10-05
Upadted ingresses, certs and R53 zones for maintenance pages

mtrbls:
The formula that worked : 

- One route53 zone per domain ( it seems we can't skip a parent when delegating) + 1:1 delegation

- No wildcard cert for domain whose parent we don't control. ( *.decisions)

- One broken or unresolvable domain prevents the whole cert from being  validated, that domain was skipped.


mogaal (dismissed):

mogaal (approved):

digitalronin (approved):
Do we need the one for 

```
maintenance-pages-demo.apps.live-1.cloud-platform.service.justice.gov.uk
```

PR #3442, opened 2020-10-06
Track a query production - add batch access to circleci

digitalali-moj:


digitalronin (changes requested):
I think "batch" is an apiGroup, not a resource. 
e.g. https://github.com/ministryofjustice/cloud-platform-environments/blob/0c53ba203f5b31bc6d026be82a6a149a5df22a53/namespaces/live-1.cloud-platform.service.justice.gov.uk/hmpps-book-secure-move-api-dev/06-serviceaccount-circleci.yaml#L43

digitalronin (changes requested):
I don't think this is the change you want to make. By changing the apiGroup from "" to "batch" you're restricting the resources on which you're granting permissions.

I think I was half-right earlier - "batch" seems to be used as both an apiGroup and a resource, e.g. 
https://github.com/ministryofjustice/cloud-platform-environments/blob/main/namespaces/live-1.cloud-platform.service.justice.gov.uk/contact-moj-development/08-service-account-circleci.yaml

I think your original change might have been correct.

Sorry, we're all working on today's incident, so I'm not able to spend as much time on PRs as usual.

digitalronin (approved):

digitalronin (approved):

PR #3443, opened 2020-10-06
PPD-47: Ensuring secure transport used for uploading documents

brightonsbox:
This reverts commit eb484fbf9a92beefaff79363039e3e34a1749e6a.


digitalronin:
Didn't #eb484fb cause the apply pipeline to fail, @brightonsbox ?

brightonsbox:
@digitalronin sorry this does reinstate that, but removes the principal block which I believe is the root cause of the issue.  Doesn't make sense to specify a principal block for an IAM policy.

digitalronin (approved):

PR #3514, opened 2020-10-07
Remove minor version number from RDS instances

digitalronin:
These RDS instances have "auto minor upgrade" enabled. This causes
problems when the minor version of the AWS instances is greater than the
minor version specified in the code.

If "auto minor upgrade" is enabled, we shouldn't specify the minor
version in the terraform code.

### DIFF
--- a/namespaces/live-1.cloud-platform.service.justice.gov.uk/formbuilder-platform-live-dev/resources/submitter.tf
+++ b/namespaces/live-1.cloud-platform.service.justice.gov.uk/formbuilder-platform-live-dev/resources/submitter.tf
@@ -11,7 +11,7 @@ module "submitter-rds-instance" {
   infrastructure-support     = var.infrastructure-support
   team_name                  = var.team_name
 
-  db_engine_version = "10.9"
+  db_engine_version = "10"
 
   providers = {
     aws = aws.london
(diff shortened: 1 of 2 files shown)


poornima-krishnasamy:
> These RDS instances have "auto minor upgrade" enabled. This causes
> problems when the minor version of the AWS instances is greater than the
> minor version specified in the code.
> 
> If "auto minor upgrade" is enabled, we shouldn't specify the minor
> version in the terraform code.

@digitalronin, not related to this PR but should we add support in our terraform RDS module to have options to disable this auto_minor_upgrade?

poornima-krishnasamy (approved):

digitalronin:
> @digitalronin, not related to this PR but should we add support in our terraform RDS module to have options to disable this auto_minor_upgrade?

Maybe. It seems like a good thing to have enabled though, so another option would be guidance (and rego policies) to have it always enabled and prevent people specifying minor versions at all. 

Let's see how this PR looks in the plan pipeline, and we can decide after we fix the current batch of problem DBs

digitalronin:
This PR is not to be merged yet - I want to see what the terraform plan output looks like, first.

digitalronin:
This is a very big **NOPE** - it would recreate the database.

PR #3535, opened 2020-10-08
Remove RDS minor version

digitalronin:
Conflicts with "auto upgrade minor"

### DIFF
--- a/namespaces/live-1.cloud-platform.service.justice.gov.uk/money-to-prisoners-prod/resources/rds.tf
+++ b/namespaces/live-1.cloud-platform.service.justice.gov.uk/money-to-prisoners-prod/resources/rds.tf
@@ -23,7 +23,7 @@ module "rds" {
 
   rds_family           = "postgres10"
   db_engine            = "postgres"
-  db_engine_version    = "10.10"
+  db_engine_version    = "10"
   db_instance_class    = "db.m5.large"
   db_allocated_storage = "50"
   db_name              = "mtp_api"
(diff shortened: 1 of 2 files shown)


mogaal (approved):

ushkarev:
@digitalronin I just spotted this change… are you confident that minor version upgrades do not affect uptime? I'm perfectly content with automatic minor version upgrades as long as they do not cause a pause in RDS connections. I haven't spotted any RDS errors, but I wasn't looking out for them because I didn't know this change was coming. Just checking as I recall an update in the past that disconnected all active connections in our Django apps, despite tests in your Rails app suggesting that it wouldn't. Might be worthwhile posting a Slack message in the ask or updates channels when such changes are made?

digitalronin:
@ushkarev This isn't actually a *change*. Your minor upgrade already happened, during the instance's regular maintenance window, so no additional downtime is incurred.

This change doesn't affect whether the minor upgrades happen or not - that's already defined in the terraform code. But, after the minor upgrade has happened, there's a conflict between the terraform code and the reality of the RDS instance. The terraform code specifies an older version than the instance, and you get an error complaining about trying to downgrade the instance (which won't happen).

ushkarev:
@digitalronin  Ok so perhaps I commented on the wrong PR and I should have posted on one of the previous module version bumps. But I don't think we _initially_ had auto minor version upgrades or did we? Given it used to be locked to 10.10 without problem?

digitalronin:
@ushkarev auto-upgrade-minor has always been the default behaviour of the [RDS module](https://github.com/ministryofjustice/cloud-platform-terraform-rds-instance) (because it's the default for AWS RDS, and the module didn't specify that it **shouldn't** happen).

I guess there just hasn't been an upgrade from 10.10 for AWS RDS since your DB was initially created (which was October 2019, so that's not unlikely).

ushkarev:
Ok I see, thanks. I know it's 10.13 now so at some point it did decide to upgrade 🤷

digitalronin:
Yeah, it'll happen whenever AWS decide to push a new minor version. You can disable that if you want to, but unless you're sure it's causing problems I wouldn't advise it, because then you miss out on all the bug/security fixes.

ushkarev:
Yup, I totally agree! You know how I worry too much :)

PR #3587, opened 2020-10-15
More memory for jpc dev

mafonso:
because Java™

### DIFF
--- a/namespaces/live-1.cloud-platform.service.justice.gov.uk/calculate-journey-variable-payments-dev/02-limitrange.yaml
+++ b/namespaces/live-1.cloud-platform.service.justice.gov.uk/calculate-journey-variable-payments-dev/02-limitrange.yaml
@@ -7,8 +7,8 @@ spec:
   limits:
   - default:
       cpu: 1000m
-      memory: 1000Mi
+      memory: 2048Mi
     defaultRequest:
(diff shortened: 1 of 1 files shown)


digitalronin (dismissed):
Ideally, we'd prefer people not use `defaultRequest` because this ring-fences cluster resources for the pod(s), *whether or not the pod uses them*.

In the past, we've had a lot of problems with resources being reserved for pods that didn't need them, leading to artificial resource shortages in the cluster.

If you're confident that at least your `defaultRequest` CPU and memory will always be used by these pods, please go ahead and merge your PR. But, if you're not sure, please leave the `defaultRequest` as is. 

The change to `default` will allow your pods to consume resources up to the limits you've specified, whenever they need them.

More info [here](https://user-guide.cloud-platform.service.justice.gov.uk/documentation/concepts/namespace-limits.html#namespace-container-resource-limits)

mafonso:
> If you're confident that at least your `defaultRequest` CPU and memory will always be used by these pods, please go ahead and merge your PR. But, if you're not sure, please leave the `defaultRequest` as is.

Good point. We are sure the the memory will always sit above that value, the same might not be true for CPU. 
So I'll revert the default to allow for cluster flexibility

digitalronin (approved):

PR #3624, opened 2020-10-19
Remove JPC staging namespace and resources

mafonso:
Cleanup deprecated namespace.


jasonBirchall (approved):
Great, thanks @mafonso

PR #3682, opened 2020-10-23
PIC-637 - Remove DynamoDB

jhevans:


digitalronin (changes requested):
You need to keep the `provider` stanza.

I'd suggest putting just that part into a `main.tf` file.

jhevans:
@digitalronin thanks for the heads up re the provider! Have updated

digitalronin:
@jhevans `terraform plan` is still failing for this - I'm not sure why. I've asked the team, and we're looking into it.

digitalronin:
@jhevans The providers for this resource were a bit messed up, so we've deleted the dynamodb instance via a manual terraform apply, and we'll merge another PR that leaves the code in this folder in a sensible state. Sorry this took a while.

jhevans:
@digitalronin ah right, thanks for letting me know, can I consider that done and applied then? Was the issue anything to do with how we've built it? If there's anything we can do to avoid such problems in future I'm always eager to learn 😄

digitalronin:
@jhevans No, it's because we forked an existing dynamodb terraform provider, which included a custom provider definition inside the module. That's not how we do things anymore, so terraform thought there was a missing provider when it tried to delete the resource, absent the code that defines the resource. We ran a `terraform destroy` targeting the dynamodb table, and everything is fine again.

We might run into the same problem if you're deleting dynamodb from staging/prod too though.

jhevans:
@digitalronin Funny you should say that 😬  - https://github.com/ministryofjustice/cloud-platform-environments/pull/3687

PR #3703, opened 2020-10-27
Update prod namespaces RDS to module v5.10

digitalronin:


digitalronin:
I've looked through the plan output. Everything looks fine. 

As well as the actual changes, there are some dummy changes where pingdom-check tags are being reordered.

The scariest thing is some elasticsearch policy changes, which I think don't actually make a real difference. ISTR that's a known issue with the elasticsearch module.

I suggest we merge the non-prod PR ASAP, and then leave it a day or two before merging this one.

pwyborn (approved):

PR #3769, opened 2020-11-06
Apply - add new permissions staging

willc-work:
add permissions required for helm3 to the service account -staging


poornima-krishnasamy:
@willc-work , i think u will need to add the RoleBinding to make the serviceAccount work. I am approving the PR but pls look into adding that if it doesnot work.

poornima-krishnasamy (approved):

poornima-krishnasamy (approved):

PR #3792, opened 2020-11-11
✨ PIC-851: Add queue and s3 bucket for crime-portal-gateway in prod

jhevans:


digitalronin (approved):
I'm not sure why concourse won't run a `plan` for this change, but I've run it manually, and it looks fine.

PR #3793, opened 2020-11-12
Add a route53 zone for prisonvisits.service.just..

digitalronin:
There is a route53 hosted zone which has an A record for a production service:
```
http://staff.prisonvisits.service.justice.gov.uk/
```
...but it's not being managed in terraform.

This change adds a terraform resource for the zone. A manual `terraform
import` will import the existing AWS route53 hosted zone into the
terraform state.


digitalronin:
Don't merge this until after the `terraform import` is done

mtrbls (approved):

digitalronin:
terraform import and apply are done, so this can be merged now

PR #3796, opened 2020-11-13
Create intervention dev namespace

sldblog:
Intervention teams are moving into the beta phase, so we need namespaces
to deploy the applications.

The namespace is "empty" as we will be adding modules later.

Split from #3795.


pwyborn (approved):

sldblog:
🎩 Thank you! 🙇‍♂️

PR #3895, opened 2020-12-03
Add scripts to post namespace usage data to HOODAW

digitalronin:
This PR refactors & renames the namespace usage script so that it runs much faster (~ 18 seconds instead of many minutes), and adds a shell script to invoke the script and post the resulting JSON to HOODAW.

This script will be invoked by a concourse pipeline job defined in another PR


jasonBirchall (approved):
This looks really sensible. We should probably add some tests to make these changes easier to review. Probably under another ticket. :)

PR #4009, opened 2020-12-31
Fix: Change apply-bot DB name

colinbruce:
I inadvertently used dashes instead of underscores


namespaces/live-1.cloud-platform.service.justice.gov.uk/laa-apply-bot-production/resources/rds.tf
```diff
@@ -18,7 +18,7 @@ module "rds" {
   infrastructure-support = "apply@digital.justice.gov.uk"
   db_engine              = "postgres"
   db_engine_version      = "11"
-  db_name                = "laa-apply-bot-production"
+  db_name                = "laa_apply_bot_production"
```
digitalronin, line 21: Underscores aren't allowed either, I'm afraid. 

The specific error from AWS is: 

> InvalidParameterValue: DBName must begin with a letter and contain only alphanumeric characters

digitalronin (changes requested):

digitalronin (approved):

PR #4021, opened 2021-01-08
start rename probation-tiering to hmpps-tier

nickmcmahon01:


jasonBirchall (changes requested):
Hi @nickmcmahon01, the namespace name has to be equal to the directory name. You'll need to change the dir name to `hmpps-tier-dev` or the metadata in your manifests to `hmpps-tier`.

nickmcmahon01:
Thanks @jasonBirchall I was looking at the variables and scratching my head 👍

jasonBirchall (approved):

PR #4045, opened 2021-01-13
Prisoner Content Hub: Add user/bucket policies to allow development S3 user to pull staging S3 content

spikeheap:
This PR enables the S3 user in the `prisoner-content-hub-development` namespace to read from the S3 bucket in `prisoner-content-hub-staging`, following the [guidance for migrating S3 buckets](https://github.com/ministryofjustice/cloud-platform-terraform-s3-bucket#migrate-from-existing-buckets).

I've hard-coded the IAM user and bucket ARNs because they're reaching across into different namespaces. If there's a way to reference it in Terraform rather than using absolute values I'd love a push in the right direction. I'll happily update this PR.


spikeheap:
There's a warning that this affects multiple namespaces, but that's intended as we're reaching out from one namespace into another, and need to make changes on both sides.

razvan-moj (approved):

PR #4052, opened 2021-01-13
Set max_allocated_storage to 1000

poornima-krishnasamy:
the ratio of iops to max storage should be at-least 0.5. This is to adjust the auto-scaling capability enabled on the module.

More details here: https://github.com/ministryofjustice/cloud-platform-environments/pull/3978

### DIFF
--- a/namespaces/live-1.cloud-platform.service.justice.gov.uk/laa-cla-backend-production/resources/rds.tf
+++ b/namespaces/live-1.cloud-platform.service.justice.gov.uk/laa-cla-backend-production/resources/rds.tf
@@ -9,7 +9,7 @@
 # Make sure you restart your pods which use this RDS secret to avoid any down time.
 
 module "cla_backend_rds" {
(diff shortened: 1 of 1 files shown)


jasonBirchall:
A Terraform plan doesn't recognise this change, it just changes Pingdom (which we know is incorrect):
https://concourse.cloud-platform.service.justice.gov.uk/builds/14960#L5ff5bfea:186:223

Would we not expect DB changes?

poornima-krishnasamy:
> A Terraform plan doesn't recognise this change, it just changes Pingdom (which we know is incorrect):
> https://concourse.cloud-platform.service.justice.gov.uk/builds/14960#L5ff5bfea:186:223
> 
> Would we not expect DB changes?

@jasonBirchall , yes there is no expected DB changes. But in order to upgrade the RDS module, if the max_allocated_storage is not set for this RDS, then it uses the default(10000) which is less than the 0.5 iops to storage ratio.

jasonBirchall (approved):

PR #4159, opened 2021-01-28
increase dynamodb message max input to solve 400 errors

andymarke:


namespaces/live-1.cloud-platform.service.justice.gov.uk/prison-to-probation-update-prod/resources/dynamodb.tf
```diff
@@ -8,7 +8,7 @@ module "message_dynamodb" {
   infrastructure-support       = var.infrastructure-support
   is-production                = "true"
   namespace                    = var.namespace
-  autoscale_max_read_capacity  = 500
+  autoscale_max_read_capacity  = 5000
   autoscale_max_write_capacity = 20
```
andymarke, line 13: with only 11,000 rows with a 10 second cron that reads some rows we reach that initial limit straight away, so going for a much bigger number to see where it settles at

razvan-moj (approved):

PR #4179, opened 2021-01-29
Restrict deletion of production database for Prisoner Money

ushkarev:


ushkarev:
am i right in thinking that this simple settings switch will _NOT_ cause any downtime or require the database to restart?

ushkarev:
i could split this PR into two… but is it really necessary? these changes are symmetrical between our 2 namespaces so would be clearer in the history if they're together 😉

jasonBirchall (approved):

ushkarev:
going to make a separate PR to check whether RDS restarts using `test` environment only

PR #4250, opened 2021-02-09
New Topic for offender assessment event

reallydontask:


namespaces/live-1.cloud-platform.service.justice.gov.uk/offender-events-prod/resources/topic.tf
```diff
@@ -68,3 +68,24 @@ resource "kubernetes_secret" "prison_data_compliance" {
...
+  topic_display_name = "offender-assessments-events"
+  providers = {
+    aws = aws.london
+  }
+}
+
+resource "kubernetes_secret" "offender_assessments_events" {
+  metadata {
+    name      = "offender-assessments-events-topic"
+    namespace = var.namespace
```
Mjwillis, line 84: Namespace looks wrong, this should be the assessment namespace - or is that not running in CP?

reallydontask: not running in CP, so it will need manual extraction as per slack thread

sablumiah (approved):

PR #4256, opened 2021-02-10
Re-adds the interventions SNS topic to domain-events

sldblog:
We're moving SNS topics to `hmpps-domain-events-{env}` namespaces so they are all discoverable in the same place.

This is part 2 of 2: adding the topic to its destination
Part 1 of 2 was #4254, 28b2ba1d340b347be652d2d349ba11e35db1524a

Related to the previous attempt of moving the topic in one go: #4252 and its revert: #4253.


digitalronin (approved):

sldblog:
🎩 🙇‍♂️ Thank you!

PR #4272, opened 2021-02-11
Add "review-after" annotation to ns template

digitalronin:
This will be populated with a date 3 months into the future, if the user
answers "yes" to the question "is this a sandbox/scratch environment?"
when running `cloud-platform environment create`

required by https://github.com/ministryofjustice/cloud-platform-cli/pull/70


digitalronin:
The conftest yaml failure is expected - this is a golang template file, but the filename says it's yaml.

poornima-krishnasamy (approved):

PR #4284, opened 2021-02-12
Redirect current interventions topic secret

sldblog:
1. Removes the `intervention-events` SNS topic.
2. Reconfigures the `intervention-events-topic` secret to point to `module.hmpps-domain-events`
3. Adds a new `hmpps-domain-events-topic` secret so it has a correct name

We're keeping the `intervention-events-topic` until the application can be reconfigured. I don't want to block work on it at the moment.


poornima-krishnasamy (approved):

namespaces/live-1.cloud-platform.service.justice.gov.uk/hmpps-domain-events-dev/resources/hmpps-interventions.tf
```diff
@@ -0,0 +1,25 @@
...
+  }
+
+  data = {
+    access_key_id     = module.hmpps-domain-events.access_key_id
+    secret_access_key = module.hmpps-domain-events.secret_access_key
+    topic_arn         = module.hmpps-domain-events.topic_arn
+  }
+}
+
+resource "kubernetes_secret" "intervention_global_events_sns" {
```
Mjwillis, line 14: @sldblog why are there two cred pushed into same namespace with same details?  Both secrets will be the same

Mjwillis (commented):
See comment

namespaces/live-1.cloud-platform.service.justice.gov.uk/hmpps-domain-events-dev/resources/hmpps-interventions.tf
sldblog: @Mjwillis We're keeping the intervention-events-topic until the application can be reconfigured. I don't want to block work on it at the moment.

sldblog:
🎩 🙇‍♂️

PR #4288, opened 2021-02-12
 PIC-871 increase defaultRequest

chris-faulkner-mojdt:


digitalronin (changes requested):
`defaultRequest` means "reserve this much resource for me, and prevent anyone else in the cluster from using it"

Unless you are absolutely certain you need to do this, it's a bad idea because it leads to the cluster running out of available resources.

This namespace is using less than half of the currently-reserved CPU for each pod, so I don't think this change is advisable:
```
$ kubectl -n court-probation-dev top pods
NAME                                         CPU(cores)   MEMORY(bytes)
court-case-matcher-58b569b8fb-hc8zk          4m           367Mi
court-case-service-54c4d6b76d-brbrs          4m           617Mi
court-case-service-54c4d6b76d-s9pvc          5m           473Mi
court-list-mock-data-6945d79466-ctn6v        1m           53Mi
crime-portal-gateway-b9d948c85-cqr9v         4m           313Mi
port-forward-pod-6fb56d74c-48qlh             0m           0Mi
prepare-a-case-75c956f5-wzvtd                1m           34Mi
probation-court-prototype-58c44cdf4f-js8mv   1m           268Mi
```
There is more information about `limit` and `defaultRequest` in [this article](https://user-guide.cloud-platform.service.justice.gov.uk/documentation/concepts/namespace-limits.html#namespace-container-resource-limits)

PR #4293, opened 2021-02-15
Remove old secret from interventions

sldblog:
https://github.com/ministryofjustice/hmpps-interventions-service/pull/98 has switched to the new one, so we no longer need to maintain the fallback name

Cleans up #4284


sldblog:
cc @Mjwillis

pwyborn (approved):

sldblog:
🎩 🙇‍♂️ Thank you!

PR #4296, opened 2021-02-15
Parliamentary questions production upgrade rds

digitalali-moj:


digitalronin (changes requested):
I would suggest keeping the skip file in place until after the upgrade has finished, or it's not doing anything useful.

digitalronin (approved):

PR #4306, opened 2021-02-16
Add route53.tf

nickmcmahon01:


jasonBirchall (changes requested):
Hi @nickmcmahon01, this looks good. I'd recommend interpolating the tag values from your [variables.tf](https://github.com/ministryofjustice/cloud-platform-environments/blob/main/namespaces/live-1.cloud-platform.service.justice.gov.uk/hmpps-tier-prod/resources/variables.tf) file rather than duplicating. [This](https://user-guide.cloud-platform.service.justice.gov.uk/documentation/other-topics/route53-zone.html#route53-tf) is a nice reference of what it should look like.

jasonBirchall (approved):
Great, thanks!

PR #4338, opened 2021-02-18
Remove unused resources - Monitoring NS

sablumiah:
Removed used resources that have either been moved to the infrastructure repo or have been superseded with a different resource


jasonBirchall (approved):
The plan looks good, it will be interesting to see how the pipeline reacts to the directory being removed i.e. will it delete the namespace?

sablumiah:
> The plan looks good, it will be interesting to see how the pipeline reacts to the directory being removed i.e. will it delete the namespace?

There are no tf resources so should be fine, also the NS yaml deletion will be ok also as the pipeline only applies what is there, if the file is removed, it is none the wiser that it was ever there.

PR #4366, opened 2021-02-24
Terraform plan and  apply to check the contents of versions.tf and select respective binary

poornima-krishnasamy:
This change allows us to run Terraform 0.12, 0.13 and 0.14 in the pipeline at the same time.


jasonBirchall (approved):
This looks great @poornima-krishnasamy :+1:

PR #4388, opened 2021-02-26
LAA CLA Frontend Staging S3 Bucket

mayowaadeniyi:
Intentionally created for public access. I heard there were cron jobs running weekly which turn the access to buckets from public to private. What do we need to do in order to be exempted from this just like how laa-cla-backend is exempted.

### DIFF
--- a/namespaces/live-1.cloud-platform.service.justice.gov.uk/laa-cla-frontend-staging/resources/s3.tf
+++ b/namespaces/live-1.cloud-platform.service.justice.gov.uk/laa-cla-frontend-staging/resources/s3.tf
@@ -1,5 +1,5 @@
 module "cla_frontend_static_files_bucket" {
-  source                        = "github.com/ministryofjustice/cloud-platform-terraform-s3-bucket?ref=4.6"
+  source                        = "github.com/ministryofjustice/cloud-platform-terraform-s3-bucket?ref=4.5"
   acl                           = "public-read"
   enable_allow_block_pub_access = false
   team_name                     = var.team_name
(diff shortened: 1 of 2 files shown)


namespaces/live-1.cloud-platform.service.justice.gov.uk/laa-cla-frontend-staging/resources/s3.tf
```diff
@@ -0,0 +1,40 @@
+module "cla_frontend_static_files_bucket" {
+  source                        = "github.com/ministryofjustice/cloud-platform-terraform-s3-bucket?ref=4.6"
```
jasonBirchall, line 2: Hey! Are you okay to drop this version down to 4.5 for now. We're in the middle of upgrading our Terraform stack. I'll update it to 4.6 on Monday next week.

jasonBirchall (changes requested):

jasonBirchall:
In regards to your questions regarding a public bucket. That's correct! We do have a job that removes the permissions of a public bucket. When the new bucket is created I can add it to our exceptions list, as outlined in our [runbook](https://runbooks.cloud-platform.service.justice.gov.uk/public-s3-bucket.html#make-an-s3-bucket-public)

namespaces/live-1.cloud-platform.service.justice.gov.uk/laa-cla-frontend-staging/resources/s3.tf
mayowaadeniyi: Done

jasonBirchall (approved):

PR #4390, opened 2021-02-26
Delete ingress.tf

davidread:
because we decided to use a shared ingress controller, after all - see: https://github.com/ministryofjustice/cloud-platform/blob/main/architecture-decision-record/019-Shared-Ingress-Controllers.md


poornima-krishnasamy (approved):

poornima-krishnasamy:
@davidread , this PR doesnot go through concourse status check. I will overwrite and merge.

PR #4505, opened 2021-03-22
Add the topic ARN to interventions research env

sldblog:
We created a separate environment to use in user research, which needs
separate configuration


mogaal (approved):

sldblog:
🎩 🙇‍♂️

PR #4582, opened 2021-04-08
Update hmpps-tier-calculation-complete.tf

nickmcmahon01:
Add secrets in a generic way usable for our tool which is deployed to multiple namespaces.


jasonBirchall (approved):
:+1:

PR #4604, opened 2021-04-13
Add policy to access Analytical Platform S3 bucket

sldblog:
Our daily data extraction needs to push files to an S3 bucket that is
currently provisioned and managed by the Analytical Platform

This change adds a policy so that our apps can access those buckets and
pushes the related secrets into the namespace

PR #4607, opened 2021-04-13
IC-1505: Add AP bucket name to the secret

sldblog:
So that the application codebase doesn't need to know how the analytical
platform bucket is configured. It can get all details from the secret

Follows up on #4604


jasonBirchall (approved):

sldblog:
🙇‍♂️ Thank you!

PR #4731, opened 2021-05-10
Upgrading the Live Production User Datastore instance and storage.

mattempty:
The platform needs upgrading to cater for more traffic and ensuring
that service can cope with increased number of concurrent users.

Co-authored-by: Brendan Butler <brendan.butler@digital.justice.gov.uk>


pwyborn (commented):
Hi @matttei Matt - can I just check - is there a reason for a big jump in pods? (could you try 100)?

brenetic:
hi @pwyborn it's not an actual figure we will definitely be using, but rather just there should we need to. 100 is also fine

brenetic:
@pwyborn have updated it to be 100

pwyborn (approved):

PR #4749, opened 2021-05-12
Concourse user for live

razvan-moj:


vijay-veeranki (commented):
I think adding one more provider for “live” in here:
https://github.com/ministryofjustice/cloud-platform-concourse/blob/main/pipelines/manager/main/build-environments/main.tf#L24-L33
and add the service account and cluster role binding in the same place?

razvan-moj:
> I think adding one more provider for “live” in here:
> https://github.com/ministryofjustice/cloud-platform-concourse/blob/main/pipelines/manager/main/build-environments/main.tf#L24-L33

the user needs to exist before the pipeline, so this will be a manual apply

razvan-moj:
moved to infra

PR #4750, opened 2021-05-12
AR-1240 upgrade rds aurora step 1

garethmdavies:
Because the RDS aurora module doesn't currently support major upgrades (see https://github.com/ministryofjustice/cloud-platform/issues/2859) this PR will remove the rds instance and then re-create it. No snapshots or data is needed to be saved as this is only a dev environment.


poornima-krishnasamy (approved):

garethmdavies:
PR not needed as will the RDS aurora module will be updated to allow major upgrades.

PR #4816, opened 2021-05-24
DT-2072: ✨ Subscribe ui to domain events too

petergphillips:


namespaces/live-1.cloud-platform.service.justice.gov.uk/offender-events-dev/resources/offender_events_ui-sub-queue.tf
```diff
@@ -111,3 +110,14 @@ resource "aws_sns_topic_subscription" "ou_events_ui_probation_subscription" {
   protocol  = "sqs"
   endpoint  = module.offender_events_ui_queue.sqs_arn
 }
+
+data "aws_sns_topic" "hmpps-domain-events" {
+  name = "cloud-platform-Digital-Prison-Services-e29fb030a51b3576dd645aa5e460e573"
```
petergphillips, line 115: had to hard code the topic arn here for domain events from that namespace - don't know if there is a better way?

razvan-moj (approved):

PR #4844, opened 2021-05-26
Use proper intervention domain in Route53 zone

sldblog:
I mixed up things in my head, the domain in the certificate definition
is the correct one


vijay-veeranki (approved):

sldblog:
🙇‍♂️ Thank you!

PR #4879, opened 2021-06-02
Create HMPPS Assess Risks and Needs (Dev)

LucasCairns:


vijay-veeranki (changes requested):
Can this be "hmpps-assess-risks-and-needs.service.justice.gov.uk" ?

So you can create ingress like "dev.hmpps-assess-risks-and-needs-dev.service.justice.gov.uk"

LucasCairns:
The issue we wanted to avoid is cookies from the `dev` subdomain polluting other subdomains such as `preprod` and `prod`, other teams have mentioned running in to the issue with using domains such as `env.namespace.service.gov.uk` - happy to change if this is a deal-breaker?

vijay-veeranki (approved):

PR #4919, opened 2021-06-08
Justicedata - added route and cert resource

sivafirst:
For prod instance.

### DIFF
--- a/namespaces/live-1.cloud-platform.service.justice.gov.uk/justicedata-prod/resources/variables.tf
+++ b/namespaces/live-1.cloud-platform.service.justice.gov.uk/justicedata-prod/resources/variables.tf
@@ -14,6 +14,10 @@ variable "namespace" {
   default = "justicedata-prod"
 }
 
+variable "domain" {
+  default = "data.justice.gov.uk"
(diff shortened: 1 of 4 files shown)


namespaces/live-1.cloud-platform.service.justice.gov.uk/justicedata-prod/resources/variables.tf
```diff
@@ -14,6 +14,10 @@ variable "namespace" {
   default = "justicedata-prod"
 }
 
+variable "domain" {
+  default = "data.justice.gov.uk"
```
vijay-veeranki, line 18: May be it needs re-name, can you please look in to 
https://ministryofjustice.github.io/technical-guidance/documentation/standards/naming-domains.html#naming-domains

poornima-krishnasamy: Had a conversation with @vijay-veeranki and @sivafirst . This is an existing domain where the nameservers will be deligated later. so the changes are fine.

PR #4973, opened 2021-06-15
Add additional serviceaccount_rules for circleci

vijay-veeranki:


jasonBirchall (commented):
Are you able to explain why we want to make these changes?

vijay-veeranki:
@jasonBirchall 
circleci is failing as I added certificate creation in the reference app deployment, and I added additional rules just in case.
https://app.circleci.com/pipelines/github/ministryofjustice/cloud-platform-reference-app/19/workflows/ac2f03d6-105d-4b21-9265-33c70f05c2cc/jobs/432

I think you are right, we don't need to create certificate in the deployment as I can add It in the reference app.

I will close this PR and raise another PR to add a certificate instead

jasonBirchall (approved):

PR #5080, opened 2021-06-30
Fix the ppud-replacement-dev-rds terraform.

dazoakley:
It looks like t2 database instances are not supported by postgresql 13.


namespaces/live-1.cloud-platform.service.justice.gov.uk/ppud-replacement-dev/resources/rds.tf
```diff
@@ -13,8 +13,8 @@ module "ppud_replacement_dev_rds" {
   rds_name          = "ppud-replacement-dev"
   rds_family        = "postgres12"
   db_engine         = "postgres"
-  db_engine_version = "13"
-  db_instance_class = "db.t2.small"
+  db_engine_version = "13.2"
```
poornima-krishnasamy, line 16: @dazoakley , can u amend to have just the major version. We have auto_minor_upgrade enabled and terraform complaints for this change when AWS does does the automatic minor upgrades.

poornima-krishnasamy (approved):

PR #5168, opened 2021-07-21
Add verification TXT record for Prisoner Money’s sendmoneytoaprisoner.justice.gov.uk domain

ushkarev:
… other changes are cosmetic to appease terraform indentation rules

PR #5169, opened 2021-07-21
Indent values so terraform doesn't complain about style

ushkarev:
Replaces #5168


poornima-krishnasamy (approved):

ushkarev:
thanks @poornima-krishnasamy!

PR #5416, opened 2021-08-25
Revert sns and sqs upgrade on cccd_* namespace. 

poornima-krishnasamy:
cccd is having an incident and getting 500. This is to rule out that the possibility of the upgrade


jasonBirchall (approved):

poornima-krishnasamy:
Closing this. No longer required as cccd incident is not related to this upgrade.

PR #5453, opened 2021-09-01
Add S3 bucket resource to Prisoner Money prod environment

ushkarev:
… and allow pods to access it via a service account that gives access to associated k8s secret


poornima-krishnasamy (approved):

ushkarev:
thanks @poornima-krishnasamy !

PR #5470, opened 2021-09-06
Fix: Remove minor version declaration

jasonBirchall:
This solves the error:

```
Error: Error modifying DB Instance cloud-platform-xxxx: InvalidParameterCombination: Cannot upgrade postgres from 12.7 to 12.5
```

As the [minor version variable](https://github.com/ministryofjustice/cloud-platform-terraform-rds-instance/blob/main/main.tf#L127) isn't set this will cause an error if Terraform tries to set it to its declared version in this file. You can't downgrade an RDS instance so this causes the above error.


razvan-moj (approved):

jasonBirchall:
Checked with the owners of the namespace, all good 👍

PR #5501, opened 2021-09-07
Create exported namespace package 

jasonBirchall:
After importing this package you can call the function
`namespace.ChangedInPR` which will return all the namespace names
changed in a Pull request.

This is useful for a number of reasons, but more specifically it allows
us to deduplicate the namespace and perform meta actions like ensuring
the PR owner is allowed to make changes to that namespaces.


pkg/namespace/namespace.go
```diff
@@ -0,0 +1,76 @@
...
+// It queries the GitHub API for all changes made in a PR. If the PR contains changes to a namespace
+// it returns a deduplicated slice of namespace names.
+func ChangedInPR(branchRef, token, repo, owner string) ([]string, error) {
+	if token == "" {
+		return nil, errors.New("You must have a valid GitHub token.")
+	}
+
+	client := githubClient(token)
+
+	str := strings.Split(branchRef, "/")
```
poornima-krishnasamy, line 27: Can you please add an example comment on how branchRef would look like so it is easy to understand the fornat

jasonBirchall: good thinking. Adding now

pkg/namespace/namespace.go
```diff
@@ -0,0 +1,76 @@
...
+	repos, _, _ := client.PullRequests.ListFiles(context.Background(), owner, repo, prId, nil)
+
+	var namespaceNames []string
+	for _, repo := range repos {
+		fmt.Println(*repo.Filename)
+		if strings.Contains(*repo.Filename, "live") {
+			// namespaces filepaths are assumed to come in
+			// the format: namespaces/live-1.cloud-platform.service.justice.gov.uk/<namespaceName>
+			s := strings.Split(*repo.Filename, "/")
+			namespaceNames = append(namespaceNames, s[3])
```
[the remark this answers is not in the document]
jasonBirchall: absolutely spot on. Thanks for catching.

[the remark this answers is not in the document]
jasonBirchall: a7c5395

pkg/namespace/namespace.go
jasonBirchall: a7c5395

poornima-krishnasamy (approved):

PR #5652, opened 2021-09-23
WFP-403 updated namespace team name in dev

carlov20:


poornima-krishnasamy:
@carlov20 , the change to the team_name, will replace all the sqs queue names in your namespace. e.g. https://concourse.cloud-platform.service.justice.gov.uk/builds/8399#L6124f249:242. Is that ok?

PR #5671, opened 2021-09-24
Adding IAM user to access laa-cla ecr repositories

said-moj:


namespaces/live-1.cloud-platform.service.justice.gov.uk/laa-cla-backend-staging/resources/iam.tf
```diff
@@ -0,0 +1,43 @@
...
+data "aws_iam_policy_document" "laa-cla-end-to-end" {
+  statement {
+    actions = [
+      "ecr:BatchCheckLayerAvailability",
+      "ecr:BatchGetImage",
+      "ecr:GetDownloadUrlForLayer",
+    ]
+
+    resources = [
+      "var.cla-public-ecr-arn"
```
Nimphal, line 22: If I understood correctly, I think you want multiple resources in here? Weren't there 3 or so repos you wanted to pull an image from? Othat than that, looks fine to me :)

said-moj: Yes there are supposed to be three here, but I am just testing with one, when that works I will add the other two

PR #5938, opened 2021-11-03
403 Prod resources for send legal mail to prisons

mikehalmamoj:


jasonBirchall:
Hi @mikehalmamoj, the "live" cluster isn't yet classified for prod workloads. It'll probably be in n weeks (think mid december), but it's going to be a rocky ride for up-time and performance. Are you happy to hold on until then?

mikehalmamoj:
> Hi @mikehalmamoj, the "live" cluster isn't yet classified for prod workloads. It'll probably be in n weeks (think mid december), but it's going to be a rocky ride for up-time and performance. Are you happy to hold on until then?

I just wanted to get the skeleton app deployed into prod - but it's not going to be used yet anyway and I don't expect any usage for at least a couple of months. It's just nice to get that pipeline forged early.

So I'm happy to hold off if that's better for you, but it should be fine if you are okay with it?

PR #5970, opened 2021-11-05
Delete hmpps-tier-dev namspace secret

poornima-krishnasamy:
The hmpps-tier-dev namespace is migrated to live cluster and no longer in live-1. 

The pipeline fails as it couldnot find/create resources between clusters. The secrets has to be copied manually.

Revert this PR back once the `hmpps-domains-events-dev` namespace is migrated to `live`

@carlov20 FYI


carlov20 (approved):
LGTM

AntonyBishop (approved):

PR #6213, opened 2021-11-25
DT-2913: 🔨 Temporarily remove copying the pre-prod secret to prod until both have been migrated

petergphillips:


petergphillips:
wrong context

PR #6305, opened 2021-12-02
Migrate pact-broker namespace to live cluster

sldblog:


sldblog:
... Right, problem with ingress annotation fixed. Entirely my fault for misreading what should've been there

sldblog:
Closing as prod namespaces are not yet being migrated

PR #6336, opened 2021-12-03
Rename `manage-recalls` resource files 

dazoakley:
No actual resource change, just renaming files so they make more sense.


jasonBirchall:
Attempting to kick off the plan pipeline.

PR #6492, opened 2022-01-06
Migrate laa-hmrc-interface-production namespace to live

stephenrichards:


vijay-veeranki:
@stephenrichards check failed for looking at right annotation for your ingress in "live-1"
https://github.com/ministryofjustice/cloud-platform-environments/runs/4728546821?check_suite_focus=true

vijay-veeranki (approved):

PR #6575, opened 2022-01-13
change to sns filter event names for prison-to-nomis-update (dev)

lbennett-moj:


lbennett-moj:
will push again shortly

PR #6619, opened 2022-01-17
Update ingress and namespace packages

poornima-krishnasamy:
- Updated ingress package adding GetAllIngressesFromCluster function
- Updated Namespace struct, Added GetAllNamespacesFromCluster, Updated function names


poornima-krishnasamy:
One of the tests GetNamespace fails because it checks a real namespace using hoodaw API. The struct of namespace is changed as part of this PR and hence json unMarshal with new struct doesnot get the values.

pkg/namespace/namespace.go
```diff
@@ -113,7 +117,7 @@ func GetNonProductionNamespaces(ns AllNamespaces) (namespaces []string, err erro
 
 // GetAllNamespaces takes the host endpoint for the how-out-of-date-are-we and
```
jasonBirchall, line 118: Can you change the comment name as well.

pkg/namespace/namespace.go
```diff
@@ -71,7 +75,7 @@ func GetNamespace(s string, h string) (Namespace, error) {
 		}
 	}
 
-	return namespace, fmt.Errorf("Namespace %s is not found in the cluster.", s)
+	return namespace, fmt.Errorf("Namespace %s is not found in the cluster", s)
```
jasonBirchall, line 78: Remove the capital in 'Namespace'

pkg/namespace/namespace.go
```diff
@@ -127,6 +131,18 @@ func GetAllNamespaces(endPoint string) (namespaces AllNamespaces, err error) {
 	return
 }
 
+// GetNamespaces takes a Kubernetes clientset and returns all namespaces with type *v1beta1.IngressList and an error.s
+func GetAllNamespacesFromCluster(clientSet *kubernetes.Clientset) ([]v1.Namespace, error) {
+
+	namespaces, err := clientSet.CoreV1().Namespaces().List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		log.Fatalln("Can'list namespaces from cluster", err.Error())
+		return nil, err
```
jasonBirchall, line 140: Can you either use the log.Fatalln oruse the err return type, not both.

poornima-krishnasamy: Updated the code

jasonBirchall (approved):

PR #6638, opened 2022-01-18
WFP-744 migrated namespace

carlov20:


namespaces/live.cloud-platform.service.justice.gov.uk/hmpps-domain-events-preprod/resources/hmpps-allocations.tf
```diff
@@ -1,10 +1,7 @@
 resource "kubernetes_secret" "hmpps-allocations" {
   metadata {
     name      = "hmpps-domain-events-topic-hmpps-allocations"
-    namespace = var.namespace
-    # Remove when namespace has been migrated
-    # name      = "hmpps-domain-events-topic"
```
petergphillips, line 6: didn't you want to uncomment the old name here?

petergphillips: we had to rename since otherwise there would be multiple secrets with the same name in the same namespace

carlov20: Thank you for spotting that, you are correct, we wanted the old name there. I've updated the PR with the old name

petergphillips (approved):
LGTM

vijay-veeranki (approved):

PR #6673, opened 2022-01-20
Migrate `money-to-prisoners-prod` namespace to `live` 

ushkarev:
This namespace used KIAM in `live-1` so several changes are required to enable IRSA instead.


ushkarev:
don't worry about that ingress weighting annotation thing, it's not actually a problem. those unannotated ingresses are just redirects for old domains and are not going to be copied across as-is to the new cluster

jasonBirchall:
Attempting to trigger the plan pipeline.

jasonBirchall (approved):

PR #6688, opened 2022-01-20
Create cjsm-dev and route53

davidread:


namespaces/live.cloud-platform.service.justice.gov.uk/cjsm-dev/resources/route53.tf
```diff
@@ -0,0 +1,25 @@
+resource "aws_route53_zone" "www_dev_cjsm_route53_zone" {
+  name = "www.dev.cjsm.justice.gov.uk"
```
vijay-veeranki, line 2: zone with the name "cjsm.justice.gov.uk" will be fine.
You can then use that and create ingress with host "dev.cjsm.justice.gov.uk"

davidread: cjsm.justice.gov.uk is hosted on GitHub Pages - it's just subdomains that I'm wanting on Cloud Platform

vijay-veeranki (approved):

PR #6702, opened 2022-01-21
Auth switch in authenticate package

vijay-veeranki:
This is to create a new function switchContextFromS3Bucket that takes the bucket name, a config file, a region and the context of a cluster and set the current context.

### DIFF
--- a/pkg/authenticate/authenticate.go
+++ b/pkg/authenticate/authenticate.go
@@ -139,14 +139,13 @@ func SwitchContextFromConfigFile(clusterCtx, kubeconfigPath string) error {
 
 // switchContextFromS3Bucket takes the bucket name, a config file, a region and a the context of a cluster and set current context
 // i.e. live/manager/live-1, and set current context in the config file.
-func SwitchContextFromS3Bucket(bucket, s3FileName, region, clusterCtx string) (err error) {
-	configFileLocation := filepath.Join("/", "tmp", "config")
+func SwitchContextFromS3Bucket(bucket, s3FileName, region, clusterCtx, kubeconfigPath string) (err error) {
 	err = KubeConfigFromS3Bucket(bucket, s3FileName, region)
 	if err != nil {
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/authenticate/authenticate.go
```diff
@@ -109,3 +109,47 @@ func CreateClientFromS3Bucket(bucket, s3FileName, region, clusterCtx string) (cl
 
 	return
 }
+
+// switchContextFromConfigFile takes a kubeconfig file and a cluster context i.e. live-1/manager/live
+// and set current context, which is usefull to switch context to multiple clusters.
```
jasonBirchall, line 114: just one "l" in useful.

pkg/authenticate/authenticate.go
```diff
@@ -109,3 +109,47 @@ func CreateClientFromS3Bucket(bucket, s3FileName, region, clusterCtx string) (cl
...
+		return fmt.Errorf("context %s doesn't exists", clusterCtx)
+	}
+
+	config.CurrentContext = clusterCtx
+	err = clientcmd.ModifyConfig(clientcmd.NewDefaultPathOptions(), config, true)
+	if err != nil {
+		return fmt.Errorf("error ModifyConfig: %w", err)
+	}
+
+	fmt.Printf("Switched to context \"%s\"", clusterCtx)
```
jasonBirchall, line 136: I'm not a big fan of print statements like this in public functions. Just a preference.

In your function here, it should just error if there's a problem. Let the caller decide if they want to output anything.

pkg/authenticate/authenticate.go
```diff
@@ -109,3 +109,47 @@ func CreateClientFromS3Bucket(bucket, s3FileName, region, clusterCtx string) (cl
 
 	return
 }
+
+// switchContextFromConfigFile takes a kubeconfig file and a cluster context i.e. live-1/manager/live
```
jasonBirchall, line 113: Capital S in SwitchContextFromConfigFile

jasonBirchall (approved):

PR #6739, opened 2022-01-24
NN-3922: removing the reference to an environment that does not yet ex…

richardpopple:
…ist. To put back later when it is created


namespaces/live.cloud-platform.service.justice.gov.uk/keyworker-api-preprod/resources/rds.tf
```diff
@@ -34,18 +34,19 @@ resource "kubernetes_secret" "dps_rds" {
 
 # This places a secret for this preprod RDS instance in the production namespace,
 # this can then be used by a kubernetes job which will refresh the preprod data.
-resource "kubernetes_secret" "dps_rds_refresh_creds" {
```
richardpopple, line 37: We don't yet have the prod namespace and so this step fails on preprod at the moment:
https://concourse.cloud-platform.service.justice.gov.uk/teams/main/pipelines/environments-live/jobs/apply-namespace-changes-live/builds/195
Commenting it out until we have a prod environment then I'll put it back in.

sablumiah (approved):

PR #6920, opened 2022-01-28
Update auth KubeConfigFromS3Bucket

vijay-veeranki:
To provide kubeConfigPath actual input supplied at function call


pkg/authenticate/authenticate.go
```diff
@@ -97,7 +97,7 @@ func CreateClientFromConfigFile(configFile, clusterCtx string) (clientset *kuber
 // Kubernetes clientset.
 func CreateClientFromS3Bucket(bucket, s3FileName, region, clusterCtx string) (clientset *kubernetes.Clientset, err error) {
 	configFileLocation := filepath.Join("/", "tmp", "config")
```
poornima-krishnasamy, line 99: @vijay-veeranki , is it ok to use the s3FileName argument when buidling the configFileLocation?

poornima-krishnasamy: so the line would be `configFileLocation := filepath.Join("/", "tmp", s3Filename)`

poornima-krishnasamy (approved):

PR #7112, opened 2022-02-16
Add interventions API healthcheck to pingdom

sldblog:
We had database connection failed errors and I would have expected a
pingdom alert but we never got any -- it was never setup 😅


vijay-veeranki (approved):

sldblog:
Thank you! 🙇‍♂️ 🎩

PR #7429, opened 2022-03-10
ESWE-280: Add Curious SQS to HMPPS-Domain-Events preprod environment

rj-adams:
Mangle filter policy with [zzz-] to prevent queue filling up with messages which will not yet be processed


Mjwillis (approved):
lgtm

poornima-krishnasamy (approved):

PR #7577, opened 2022-03-23
Upgrade postgres to 13.6

laurentqro:
Upgrade postgres to 13.6


poornima-krishnasamy:
@laurentqro , can u push an empty commit to trigger the concourse pipeline please

PR #7786, opened 2022-04-20
Incentives prototype: Disable override of Dockerfile

xoen:
as suggested in #ask-cloud-platform, to allow prototype container to use
newer version of Node.js base image.


poornima-krishnasamy (approved):

xoen:
Thanks @poornima-krishnasamy ☺️

PR #7890, opened 2022-05-09
Remove ingress checker module

jasonBirchall:
Overview
---

We've migrated so we don't need to confirm users have updated their ingress resources. This code can live forever in the ether of GitHub's archive.


pwyborn (approved):

jasonBirchall:
Tests fail that actually have nothing to do with this PR. Closing this issue until we can resolve.

PR #7936, opened 2022-05-13
WFP-1051 added hmpps workload queue

carlov20:


jasonBirchall:
Hi @carlov20 - I can see you have made a number of amendments to this namespace. Is it worth having you added to the `dps-tech` GitHub team (I can't do this, BTW ;)). Perhaps add a team you are to the namespace and have it approved by the `dps-tech` team?

carlov20:
I've requested a review from the `probation-integration` team as my team has discussed the change with them before so they are the most appropriate to review.

While I have made a number of amendments in the namespace the team I'm part of doesn't own this domain, we are just heavy users of offender events

mxco86 (approved):
👍

razvan-moj (approved):

PR #8014, opened 2022-06-07
Add SED Restricted Team to RBAC for SOC

alex-hedley:
Add Secure Estate Digital Team Restricted Team (SED) to RBAC for SOC PreProd


petergphillips (approved):
LGTM

PR #8057, opened 2022-06-14
Alert if slow responses rate spikes up

sldblog:
To alert on spikes like this:

image


sldblog:
[Need alerting on an increased volume of slow responses](https://trello.com/c/WrAvDuBS/401-need-alerting-on-an-increased-volume-of-slow-responses)

PR #8175, opened 2022-07-01
Set up 5xx error alerts for prod and nonprod

jackmaloney:


razvan-moj (approved):

razvan-moj (approved):

razvan-moj (changes requested):
namespace: make-recall-decision-preprod
 is not the current namespace, this will not work

razvan-moj (approved):

PR #8177, opened 2022-07-01
Add offender-management-test environment for Manage POM Cases

ayqazi:
## Purpose

This is the first PR for a new environment is to allow us to test WIP code on branches using the same synthetic data as our staging environment. This allows us to test WIP code with richer data than is available on our local machines without requiring security clearance to use live data.


namespaces/live.cloud-platform.service.justice.gov.uk/offender-management-test/00-namespace.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: offender-management-test
+  labels:
+    cloud-platform.justice.gov.uk/is-production: "true"
```
jackstockley89, line 6: could this please be changed to 
`cloud-platform.justice.gov.uk/is-production: "false"`

namespaces/live.cloud-platform.service.justice.gov.uk/offender-management-test/resources/variables.tf
```diff
@@ -0,0 +1,19 @@
+variable "environment-name" {
+  default = "test"
+}
+
+variable "team_name" {
+  default = "offender-management"
+}
+
+variable "is-production" {
```
jackstockley89, line 9: could this please be changed to 
`variable "is_production" {
  default = "false"
}`

namespaces/live.cloud-platform.service.justice.gov.uk/offender-management-test/00-namespace.yaml
ayqazi: addressed

jackstockley89 (approved):
LGTM

ayqazi:
Converting to draft for 2 reasons:

1) I have to ensure the service is hooked up to the right domain. It's not a standard hosting so I can't just look at the docs
2) It has to connect to the staging DB - the current IaC creates its own DB

PR #8227, opened 2022-07-14
INC-479: create queue in preprod for incentives api to subscribe to prisoner domain events

neilmendum:
https://dsdmoj.atlassian.net/browse/INC-479


sj-williams (approved):
Great, approved.

PR #8281, opened 2022-07-25
PI-340 Add GitHub Actions runner for hmpps-probation-integration-services

marcus-bcl:


jasonBirchall:
@marcus-bcl Just discussing this with the team - the compute request is almost a third of a node, so we need to figure out how best to proceed (it should be fine).

I'm loving the use of a runner in the cluster though 👍

marcus-bcl:
Thanks @jasonBirchall, glad it's sparking a discussion 😄 

The concerns about sizing could be alleviated longer-term by implementing webhook-driven scaling with scale-to-zero enabled - which would mean the runner is only started when a job is running. There are some existing implementations of this (see https://github.com/actions-runner-controller/actions-runner-controller#autoscaling-tofrom-0) but I hit permissions errors when trying to use it in the cluster.

Alternatively for now I'd happily reduce the specs for our repo's runner.  I doubt we need the full 2CPUx7GB for what we're doing, I just copied what was used on the GitHub-hosted runners.

marcus-bcl:
I've removed the `resources` section so that it uses the default sizing.  After testing, our CI job runs in roughly the same amount of time - so we don't need the extra resource.

jasonBirchall:
Okay, brill! That makes it easier to get over the line. Let's merge it in and see how it fairs.

jasonBirchall (dismissed):

jasonBirchall:
If this works, I wouldn't mind giving this a go with some of the platform stuff (although, thinking about it, that might need hosting in a different cluster)

marcus-bcl:
@jasonBirchall could I get another approval?  I've moved the image into a ministryofjustice repository.

jasonBirchall (approved):

PR #8283, opened 2022-07-25
Prisoner content hub - prod - create skip file for RDS upgrade

leonk:
Create a skip file so that we can apply changes from https://github.com/ministryofjustice/cloud-platform-environments/pull/8282


pwyborn (approved):

leonk:
Not required.

PR #8369, opened 2022-08-15
LAA-Apply-slack-bot: Upgrade RDS

colinbruce:
I understand this process will involve downtime and that is not a problem for this service


poornima-krishnasamy (approved):

poornima-krishnasamy (commented):
@colinbruce , The plan shows the parameter group is still in postgress 11. Can you add a skip to the pipeline and amend the parameter group so the upgrade creates a new parameter group: https://user-guide.cloud-platform.service.justice.gov.uk/documentation/other-topics/upgrade-rds-db.html#disadvantages

colinbruce:
Tnanks @poornima-krishnasamy - that is odd, I _did_ add the skip-file but it got missed from the commit 🤔 
I'll re-add and re-push :+1:

PR #8391, opened 2022-08-17
Tagging resources for Read Only access to the AWS Console

mattempty:
Adding default tag globally as per documentation. THis will allow read only access to the AWS console for resources in this namespace.


mattempty:
Needs fixing errors raised.

PR #8543, opened 2022-09-01
Adding new prisoner-content-hub-db-backups namespace.

leonk:
This adds a new namespace `prisoner-content-hub-db-backups`.  This will contain an s3 bucket to store database backups of the `prisoner-content-hub-production` database.  (There will be no other services running in this namespace).
Backups will be created using a cronjob, in `.sql` format.

**Why not use the `prisoner-content-hub-production` namespace?**
The reason for using a separate namespace, is that we want dev and staging to also be able to connect to the s3 bucket, in order to run automated database refresh jobs, syncing the db with production.  
If we placed the s3 bucket in production, we would have to expose the creds for this onto non-prod environments.  Having a separate namespace is much cleaner.

Note that there is no sensitive prisoner data in this application, just content added in Drupal.

**Whatabout RDS snapshots?**
We are aware that we have RDS snapshots enabled by default, so we already have backups there.
However, we would be unable to automatically restore the db between environments from an RDS snapshot.
We also would benefit from a secondary backup option, that stores backups in a format that is easier to inspect and restore from.

### DIFF
--- a/namespaces/live.cloud-platform.service.justice.gov.uk/prisoner-content-hub-db-backups/resources/s3.tf
+++ b/namespaces/live.cloud-platform.service.justice.gov.uk/prisoner-content-hub-db-backups/resources/s3.tf
@@ -31,7 +31,7 @@ module "db_backups_storage" {
   ]
 }
 
-resource "kubernetes_secret" "db_backups_secret" {
+resource "kubernetes_secret" "db_backups_s3_production" {
   metadata {
     name      = "db-backups-s3"
     namespace = prisoner-content-hub-production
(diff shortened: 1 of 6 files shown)


namespaces/live.cloud-platform.service.justice.gov.uk/prisoner-content-hub-db-backups/resources/s3.tf
```diff
@@ -0,0 +1,75 @@
...
+    namespace = "prisoner-content-hub-development"
+  }
+
+  data = {
+    access_key_id     = module.db_backups_storage.access_key_id
+    secret_access_key = module.db_backups_storage.secret_access_key
+    bucket_arn        = module.db_backups_storage.bucket_arn
+    bucket_name       = module.db_backups_storage.bucket_name
+  }
+}
```
leonk, line 75: This adds the s3 credentials as a k8ns secret to each of the `prisoner-content-hub` environments.

The alternative would be to use bucket policy rules, to allow the s3 user from each environment to have access to this s3 bucket.  But I think the way above is better, because the s3 users on each environment are related to the buckets storing content (audio/video/images) and nothing to do with database backups.

sj-williams (approved):

PR #8554, opened 2022-09-02
OFFENCES-114 Delete old postgres 10 database and replace with postgres 14

yasinmustafa:


namespaces/live.cloud-platform.service.justice.gov.uk/manage-offences-api-dev/resources/rds.tf
```diff
@@ -1,34 +1,37 @@
...
-  source                 = "github.com/ministryofjustice/cloud-platform-terraform-rds-instance?ref=5.16.11"
-  cluster_name           = var.cluster_name
-  team_name              = var.team_name
-  business-unit          = var.business_unit
-  application            = var.application
-  is-production          = var.is_production
-  namespace              = var.namespace
-  environment-name       = var.environment
-  infrastructure-support = var.infrastructure_support
+module "manage_offences_rds" {
```
yasinmustafa, line 1: not sure - but I think I had to change the module name? so the old DB instance gets deleted? 
then a new DB will be created with this definition?

vijay-veeranki (approved):

PR #8633, opened 2022-09-12
[DSE-607] generate random passwords for prototypes

mpbostock:
this ensures that the password set in the kube cluster is in sync with the IaC


pwyborn (approved):

Burdis (commented):
lgtm

PR #8811, opened 2022-10-04
Create new PostgreSQL14 instance on prod

sldblog:
We are running on PostgreSQL 10, which will be end-of-life soon.

This PR creates a new PostgreSQL 14 instance, so we have something to migrate to.


vijay-veeranki (approved):

sldblog:
[Update postgres to latest CP supported version as 10 will be dead in November 2022](https://trello.com/c/vge330Xk/320-update-postgres-to-latest-cp-supported-version-as-10-will-be-dead-in-november-2022)

PR #8816, opened 2022-10-04
Prisoner content hub production - Upgrading redis instance size.

leonk:
We are seeing Redis errors such as 
```
OOM command not allowed when used memory > ‘maxmemory’
```
The error relates to the memory limit in Redis being reached. See https://ma.ttias.be/redis-oom-command-not-allowed-used-memory-maxmemory/


poornima-krishnasamy (approved):

leonk:
No longer required. OOM error was due to an application error, not handling a large text blob correctly.

PR #8835, opened 2022-10-10
Increase memory on interventions-preprod

sldblog:
To match change to prod in cdfd39925fb78f2b3f118ec6e2c6566d578e831f as we have the same data size on preprod


sj-williams (approved):

sldblog:
Thank you! 🎩

PR #8879, opened 2022-10-17
MAS-444: Use new event type in tier calc complete queue

sp-watson:


sp-watson:
We have found that this is actually the wrong approach - we will submit a new PR

PR #9113, opened 2022-11-04
HEAT-16: add GithubTeam to hmpps-template-kotlin

neilmendum:
https://dsdmoj.atlassian.net/browse/HEAT-16

### DIFF
--- a/namespaces/live.cloud-platform.service.justice.gov.uk/hmpps-template-kotlin/resources/main.tf
+++ b/namespaces/live.cloud-platform.service.justice.gov.uk/hmpps-template-kotlin/resources/main.tf
@@ -5,15 +5,35 @@ terraform {
 
 provider "aws" {
   region = "eu-west-2"
+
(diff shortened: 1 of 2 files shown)


namespaces/live.cloud-platform.service.justice.gov.uk/hmpps-template-kotlin/resources/main.tf
```diff
@@ -5,15 +5,35 @@ terraform {
 
 provider "aws" {
   region = "eu-west-2"
+
+  default_tags {
+    tags = {
+      # see https://user-guide.cloud-platform.service.justice.gov.uk/documentation/getting-started/accessing-the-cloud-console.html
```
neilmendum, line 11: Adding this comment because this is a template namespace for HMPPS - teams may use it for reference and they may not know what the param is used for

sj-williams (approved):

PR #9186, opened 2022-11-09
Add tags for workload SQS queues

sp-watson:


petergphillips (approved):
LGTM

sp-watson:
The syntax is wrong and there is no obvious way to fix it - this needs more research!

PR #9244, opened 2022-11-14
Check if the ingress belong to the same namespace

poornima-krishnasamy:


jasonBirchall:
@poornima-krishnasamy do you want to make this a draft so it doesn't appear in our daily reports?

PR #9247, opened 2022-11-15
NN-4479: temporarily bump whereabouts-api-preprod/rds to db.t3.large

neilmendum:
https://dsdmoj.atlassian.net/browse/NN-4479


namespaces/live.cloud-platform.service.justice.gov.uk/whereabouts-api-preprod/resources/rds.tf
```diff
@@ -4,18 +4,20 @@ variable "cluster_name" {
...
+  vpc_name                    = var.vpc_name
+  team_name                   = var.team_name
+  business-unit               = var.business-unit
+  application                 = var.application
+  is-production               = var.is-production
+  namespace                   = var.namespace
+  environment-name            = var.environment-name
+  infrastructure-support      = var.infrastructure-support
+  allow_major_version_upgrade = "false"
+  db_instance_class           = "db.t3.large"
```
neilmendum, line 18: Currently it's on `db.t2.small`. Bumping because
- `t3` is compatible with Postgres 14
- `large` should speed up the time it takes to upgrade

Will drop back to `db.t4g.small` after the upgrade

jakemulley (approved):

PR #10301, opened 2023-01-23
Refactor: Extract default tags local for HMPPS Integration API

chubberlisk:
- Extract default tags local.
- Remove unnecessary comments in ECR files.


chubberlisk:
Will split this into separate PRs!

PR #10462, opened 2023-01-25
INC-964: Increase Incentives prod DB instance class

xoen:
Changed DB instance class to `db.t4g.large`.

Also:
- added deletion protection
- added backup/maintenance window
- added explicit `allow_minor_version_upgrade = true` (this was already the case because of the [TF module default](https://github.com/ministryofjustice/cloud-platform-terraform-rds-instance#input_allow_minor_version_upgrade) for this value)

### Downtime
The [module has `apply_immediately = true `](https://github.com/ministryofjustice/cloud-platform-terraform-rds-instance/blob/64a95b9785d72009ec979a52053cd1400effa8c8/main.tf#L110) which means once this is merged the change will be applied straight away.

The backup/maintenance window are added for other/future automated changes (e.g. minor version upgrades or patching)

The [module configures the database as a Multi-AZ](https://github.com/ministryofjustice/cloud-platform-terraform-rds-instance/blob/64a95b9785d72009ec979a52053cd1400effa8c8/main.tf#L124) DB instance deployment.
This means that the downtime should be limited to the time of the failover, possibly a minute or two.
[See AWS RDS documentation](https://aws.amazon.com/premiumsupport/knowledge-center/rds-mysql-downtime-impact/#Updating_the_DB_instance_class) (page is for MySQL but process should be the same)

#### Multi-AZ deployment
In a Multi-AZ deployment AWS keeps another DB instance in standby in a different Availability Zone (AZ).
This is not normally used but is kept in sync with the Primary.

If there is a problem with the Primary instance/AZ the Standby can "failover" and become the Primary. This can take a few seconds/minutes and this can cause downtime.

The same approach is used when changing the instance class:
1. The Standby instance is modified first (no downtime as this is not used)
2. The Standby causes a "failover" becoming the Primary instance (during this switch there is a brief downtime)
3. The old Primary instance is now the Standby instance (not used)
4. The change is applied to this instance too

### DIFF
--- a/namespaces/live.cloud-platform.service.justice.gov.uk/hmpps-incentives-prod/resources/rds.tf
+++ b/namespaces/live.cloud-platform.service.justice.gov.uk/hmpps-incentives-prod/resources/rds.tf
@@ -9,9 +9,10 @@ module "dps_rds" {
   environment-name            = var.environment
   infrastructure-support      = var.infrastructure_support
   rds_family                  = var.rds_family
-  allow_major_version_upgrade = "false"
   db_instance_class           = "db.t3.small"
   db_engine_version           = "14"
+  allow_major_version_upgrade = "false"
+  allow_minor_version_upgrade = "true"
 
   backup_window          = var.backup_window
   maintenance_window     = var.maintenance_window
--- a/namespaces/live.cloud-platform.service.justice.gov.uk/hmpps-incentives-prod/resources/rds.tf
+++ b/namespaces/live.cloud-platform.service.justice.gov.uk/hmpps-incentives-prod/resources/rds.tf
@@ -9,7 +9,7 @@ module "dps_rds" {
   environment-name            = var.environment
   infrastructure-support      = var.infrastructure_support
   rds_family                  = var.rds_family
-  db_instance_class           = "db.t3.small"
+  db_instance_class           = "db.t4g.large"
   db_engine_version           = "14"
   allow_major_version_upgrade = "false"
   allow_minor_version_upgrade = "true"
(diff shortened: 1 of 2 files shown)


namespaces/live.cloud-platform.service.justice.gov.uk/hmpps-incentives-prod/resources/rds.tf
```diff
@@ -9,9 +9,14 @@ module "dps_rds" {
   environment-name            = var.environment
   infrastructure-support      = var.infrastructure_support
   rds_family                  = var.rds_family
-  allow_major_version_upgrade = "false"
-  db_instance_class           = "db.t3.small"
+  db_instance_class           = "db.t3.large"
```
xoen, line 12: Igor suggests we may well use the [newer instances](https://aws.amazon.com/rds/instance-types/) `db.t4g.large`.

ushkarev: …because it's a bit better performant and cheaper. plus, it's likely to have a longer pg version upgrade path

xoen: FYI: Done here: https://github.com/ministryofjustice/cloud-platform-environments/pull/10462/commits/4a573f9b79987523e8fa1e9fff2b46f3bf81d0ba

namespaces/live.cloud-platform.service.justice.gov.uk/hmpps-incentives-prod/resources/variables.tf
```diff
@@ -49,6 +49,12 @@ variable "number_cache_clusters" {
   default = "2"
 }
 
-variable "rds_family" {
-  default = "postgres14"
```
xoen, line 53: FYI: @ushkarev I inlined this variable as discussed as this value goes with the `db_engine_version` one.

poornima-krishnasamy (approved):

PR #10534, opened 2023-01-31
Delete old snapshots

vijay-veeranki:
Related to:
https://github.com/ministryofjustice/cloud-platform/issues/4073

This script will be scheduled to run every day in a concourse pipeline to delete snapshots older than given days.


cmd/delete-oldsnapshots/delete-oldsnapshots.go
```diff
@@ -0,0 +1,62 @@
...
+	output, err := svc.DescribeSnapshots(&ec2.DescribeSnapshotsInput{Filters: []*ec2.Filter{
+		{
+			Name: aws.String("owner-id"),
+			Values: []*string{
+				aws.String(*ownerId),
+			},
+		},
+	}})
+	if err != nil {
+		log.Fatal("Error to describe snapshots:", err.Error())
```
jaskaransarkaria, line 32: Instead of logging here is it worth returning the error and logging from the caller?

cmd/delete-oldsnapshots/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM golang:1.19
```
jaskaransarkaria, line 1: As you are running the binary it might be worth seeing if you can run this in a scratch container. This image build currently comes to ~1.54GB, it would be significantly smaller and also more secure if that were run from a container built from "scratch". Happy to pair with you on this :)

jaskaransarkaria (approved):

jaskaransarkaria (approved):

PR #10536, opened 2023-01-31
added certs with dns names

chidionyemaHMCTS:


sj-williams (dismissed):

vijay-veeranki (changes requested):
Can you please create a single zone to create certificates please

vijay-veeranki (approved):

PR #9685, opened 2023-01-06
remove_athena_reporting_s3_bucket_from_pathfinder_dev_force

jonhindle:
Remove athena resource, reporting s3 bucket

PR #10579, opened 2023-02-06
Remove Athena from pathfinder dev (#9669)

jakemulley:
This PR replays #9669 as `aws_athena_database` needed to have `force_destroy = true` set; and reverts #9708 as #9669 blocked the apply pipeline.

Needs approval from @jonhindle before merge; and then #9685 can be merged after this too.


jonhindle (approved):
lgtm

PR #10294, opened 2023-01-20
feat: 🎸 add ecr repo

jaskaransarkaria:
- add ecr repo

Planning to manually deploy the image to ecr as the repo already has ci/cd, so there's no need to create a service account or github actions

PR #10595, opened 2023-02-06
feat: 🎸 add github action to check terrform module versions

jaskaransarkaria:
When PRs add new terraform modules, then it is important that the user is using the most up-to-date module version. The github action will fail if they aren't using the latest version.

To test, you can:

1) run `go test ./...` in `cmd/check-terraform-modules-are-latest/`

Or

2) build the image and run the action against an existing pr, here's an example pr which fails as it is using an old module version:

```bash
cd cmd/check-terraform-modules-are-latest
docker build -t gha-check-terraform-version . 
docker run --rm=true --env="REPO_NAME=cloud-platform-environments" --env="API_URL=https://modules.apps.live-2.cloud-platform.service.justice.gov.uk/" --env="PR_NUMBER=10294" gha-check-terraform-version
```

[This is the PR being referenced](https://github.com/ministryofjustice/cloud-platform-environments/pull/10294)


.github/workflows/check-terraform-module-version.yaml
```diff
@@ -0,0 +1,18 @@
...
+    types: [opened, edited, reopened, synchronize]
+
+jobs:
+  format-code:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Check for editted files for terraform modules
+        uses: ministryofjustice/cloud-platform-environments/cmd/check-terraform-modules-are-latest@add-gh-action-to-pin-terraform-modules-to-latest
```
jaskaransarkaria, line 12: This will be changed to `@main` in a following PR

vijay-veeranki (approved):

PR #10601, opened 2023-02-07
ci: 🎡 activate gh action to check terraform module versions

jaskaransarkaria:


.github/workflows/check-terraform-module-version.yaml
```diff
@@ -8,8 +8,8 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      - name: Check for editted files for terraform modules
-        uses: ministryofjustice/cloud-platform-environments/cmd/check-terraform-modules-are-latest@add-gh-action-to-pin-terraform-modules-to-latest 
+      - name: Check editted files for terraform modules
```
jasonBirchall, line 11: Small change: `Check edited files for terraform modules`

.github/workflows/check-terraform-module-version.yaml
```diff
@@ -8,8 +8,8 @@ jobs:
...
+        uses: ministryofjustice/cloud-platform-environments/cmd/check-terraform-modules-are-latest@main
```
jasonBirchall, line 12: Do we want to pin the version?

.github/workflows/check-terraform-module-version.yaml
jaskaransarkaria: Thanks!

.github/workflows/check-terraform-module-version.yaml
jaskaransarkaria: Given that it is only our team using the action and currently we are only using it within this repository at the minute, I think it's ok to have it point at 'main'

.github/workflows/check-terraform-module-version.yaml
```diff
@@ -8,11 +8,11 @@ jobs:
...
       - uses: actions/checkout@v3
 
-      - name: Check for editted files for terraform modules
-        uses: ministryofjustice/cloud-platform-environments/cmd/check-terraform-modules-are-latest@add-gh-action-to-pin-terraform-modules-to-latest 
+      - name: Check edited files for terraform modules
+        uses: ministryofjustice/cloud-platform-environments/cmd/check-terraform-modules-are-latest@main
         env:
           REPO_NAME: cloud-platform-environments
-          API_URL: ${{ env.TERRAFORM_MODULE_VERSIONS_API_URL }}
+          API_URL: ${{ vars.TERRAFORM_MODULE_VERSIONS_API_URL }}
```
jaskaransarkaria, line 15: Previously, we couldn't access the env var as we were using the wrong context, to access configuration variables we need to use the vars context.

jasonBirchall (approved):

PR #10622, opened 2023-02-07
Remove unnecessary username from redis url

jrmhaig:
The latest major version of the redis gem requires that the dummy username is not present in the url. CCCD currently has a fix that will strip it out but this fix will be removed once the urls are fixed here.


namespaces/live.cloud-platform.service.justice.gov.uk/cccd-dev-lgfs/resources/elasticache.tf
```diff
@@ -30,7 +30,7 @@ resource "kubernetes_secret" "cccd_elasticache_redis" {
     primary_endpoint_address = module.cccd_elasticache_redis.primary_endpoint_address
     auth_token               = module.cccd_elasticache_redis.auth_token
     member_clusters          = jsonencode(module.cccd_elasticache_redis.member_clusters)
-    url                      = "rediss://dummyuser:${module.cccd_elasticache_redis.auth_token}@${module.cccd_elasticache_redis.primary_endpoint_address}:6379"
```
jakemulley, line 33: You'll want to retain the `:` to signify a difference between username (which will then be blank) and password.

jakemulley (approved):

PR #10831, opened 2023-02-17
Turn off github-actions-no-plain-text-action-secrets check, as needs refining

jakemulley:
This check is causing some false positives when using a `GitHub` resource, so this turns it off for now whilst we refine it.


jaskaransarkaria (approved):
Does the comma at the end break anything? LGTM otherwise

jakemulley:
@jaskaransarkaria it doesn't break anything, but good spot: I'll remove it 👍🏼

jaskaransarkaria (approved):

PR #10908, opened 2023-02-20
Export more RDS outputs in prisoner money test namespace

ushkarev:


ushkarev:
The RDS module read-me suggests that cluster_name is required but I notice that most namespaces do not include it: _is_ it required??

https://github.com/ministryofjustice/cloud-platform-terraform-rds-instance/blob/b845fbcf53d6c71dc3caa7b17dd9b478367d0054/README.md?plain=1#L79

vijay-veeranki (approved):

PR #10975, opened 2023-02-22
feat(cmd): Resource namespace check for Pull Requests

jackstockley89:
Script covers all resources where namespaces are stated in the metadata.

future work:
- add output of the resources effected to the comment
- add section to review modules as well 
- fix output as it is duplicates each mismatch


.github/workflows/compare-namespaces.yaml
```diff
@@ -0,0 +1,49 @@
...
+env:
+  GITHUB_OAUTH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+jobs:
+  compare-namespaces:
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
```
jaskaransarkaria, line 19: You don't need to use matrix here, because you are only running it on `ubuntu-latest` you can get rid of the strategy bit and simplify it to:

```
runs-on: ubuntu-latest
```

cmd/compare-namespace/compare-namespace.go
```diff
@@ -0,0 +1,218 @@
...
+				}
+			}
+		}
+	}
+	return varNamespace, nil
+}
+
+func main() {
+	if *token == "" {
+		client = github.NewClient(nil)
```
jaskaransarkaria, line 171: Is it worth logging here that we are init-ing a client without a token that could hit github rate limits?

jaskaransarkaria (approved):
LGTM

jaskaransarkaria (approved):

jaskaransarkaria (approved):

cmd/compare-namespace/compare-namespace.go
```diff
@@ -0,0 +1,272 @@
...
+
+	if diags.HasErrors() {
+		return nil, fmt.Errorf("error getting TF resource: %s", diags)
+	}
+	blocks = f.Body().Blocks()
+
+	return blocks, nil
+}
+
+// resouceType will search for namespace in all resources raised in a Pull Request
```
jaskaransarkaria, line 93: // resourceType

jaskaransarkaria, line 93:
```suggestion
// resource type will search for namespace in all resources raised in a Pull Request
```

PR #10991, opened 2023-02-22
Enables restore/backup from S3 bucket

andrewl:


namespaces/live.cloud-platform.service.justice.gov.uk/make-recall-decision-db-analysis/resources/rds-mssql.tf
```diff
@@ -36,6 +37,24 @@ module "rds_mssql" {
   # This will rotate the db password. Update the value to the current date.
   # db_password_rotated_date  = "dd-mm-yyyy"
 
+	resource "aws_db_option_group" "ppud_backup_rds_option_group" {
```
poornima-krishnasamy, line 40: can u please move the resource block out of the module

PR #11202, opened 2023-03-03
live, all namespaces, versions integrity: set required_providers

jakemulley:
This PR does three things to the namespaces:

- adds missing required_providers to namespaces
- adds missing versions to required_providers
- removes unused providers from required_providers


jaskaransarkaria (approved):
LGTM

PR #11214, opened 2023-03-06
Rotate rds and elasticache secrets (DASO, VCMS and JitBit dev envs)

silvianahorga-al:


namespaces/live.cloud-platform.service.justice.gov.uk/hmpps-victim-case-management-system-dev/resources/redis.tf
```diff
@@ -1,16 +1,17 @@
 module "elasticache_redis" {
-  source                 = "github.com/ministryofjustice/cloud-platform-terraform-elasticache-cluster?ref=5.5"
-  vpc_name               = var.vpc_name
-  application            = var.application
-  environment-name       = var.environment
-  is-production          = var.is_production
-  infrastructure-support = var.infrastructure_support
-  team_name              = var.team_name
-  number_cache_clusters  = "2"
-  node_type              = "cache.t2.small"
```
adeweetman-al, line 10: Looks good, only thing might be to change the redis type to cache.t3.small, to use a more modern type as discussed.

silvianahorga-al: changed it now

sj-williams (approved):

PR #11399, opened 2023-03-09
Remove unused pathfinder- variables

jakemulley:
I compared these with the direct variables (e.g. `variable "is_production"`) and they're the same.


jaskaransarkaria (approved):
Deleting `is-production` variable makes me feel nervous, I'm sure it's fine but I did something similar on bichard and didn't realise the value was used somewhere else and that caused terraform to start deleting prod 🙈

PR #11667, opened 2023-03-20
Route53 config for find-unclaimed-court-money

vertism:


poornima-krishnasamy:
Hi @vertism , is the zone moved to the new namespace? Is this namespace still required?

vertism:
> Hi @vertism , is the zone moved to the new namespace? Is this namespace still required?

Hi @poornima-krishnasamy, it is still required. We're waiting on sign off from stakeholders, but that should be soon.

jakemulley (approved):

PR #11866, opened 2023-03-30
LEP-39 CFE Civil Staging. Re-generate service account

starswan:
CFE Civil Staging appears to have a faulty service account. 

This is the result of re-generating using cloud-platform environment serviceaccount create with cloud-platform-cli 1.26.1


davidread:
we don't need this any after all

PR #12161, opened 2023-04-18
Add namespace for CAS-2 dev environment

Harriethw:
Add a namespace and ECR for Community Accommodation Service 2 dev


Harriethw:
joining existing cas namespace instead

PR #12525, opened 2023-04-25
CRIMRE-280 Create crime review production namespace

zheileman:
Also adding the RDS resource at the same time, not sure if it is best to do it separately?


poornima-krishnasamy (approved):

poornima-krishnasamy:
> Also adding the RDS resource at the same time, not sure if it is best to do it separately?

Its ok. You can add RDS together when creating the namespace

PR #12795, opened 2023-05-03
HAAR-1392: enable cross-namespace traffic routing between client serv…

ravmoj:
…ices and auth

### DIFF
--- a/namespaces/live.cloud-platform.service.justice.gov.uk/hmpps-audit-dev/04-networkpolicy.yaml
+++ b/namespaces/live.cloud-platform.service.justice.gov.uk/hmpps-audit-dev/04-networkpolicy.yaml
@@ -24,4 +24,4 @@ spec:
     - from:
         - namespaceSelector:
             matchLabels:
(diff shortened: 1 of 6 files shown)


namespaces/live.cloud-platform.service.justice.gov.uk/hmpps-audit-dev/04-networkpolicy.yaml
```diff
@@ -24,4 +24,4 @@ spec:
     - from:
         - namespaceSelector:
             matchLabels:
-              component: ingress-controllers
+              component: hmpps-audit-dev
```
poornima-krishnasamy, line 27: This would stop access the auth url from the browser or from any other service. Can you create a new network policy and allow a target-namespace. Also the label should be: `cloud-platform.justice.gov.uk/namespace: "source-namespace'` where the source-namespace is from where the request is coming from.

poornima-krishnasamy (approved):

PR #13119, opened 2023-05-16
LEP-119 Remove partner directories (check-financial-eligibility-partner-*)

ripan:


ripan:
Need to create a separate PR for each namespace.

PR #13569, opened 2023-06-02
NN-5154: configure IRSA for whereabouts-api-dev

neilmendum:
https://dsdmoj.atlassian.net/browse/NN-5154


namespaces/live.cloud-platform.service.justice.gov.uk/whereabouts-api-dev/resources/main.tf
```diff
@@ -4,7 +4,7 @@ terraform {
 }
 
 provider "aws" {
-  region = "eu-west-1"
+  region = "eu-west-2"
```
neilmendum, line 7: Set default region to match other resources, as was not able to find SSM param in `eu-west-1`

jakemulley (approved):

PR #13574, opened 2023-06-02
NN-5154: configure IRSA for whereabouts-api-preprod

neilmendum:
https://dsdmoj.atlassian.net/browse/NN-5154


neilmendum:
Made changes in wrong namespace, will raise new PR

PR #13630, opened 2023-06-06
Correcting extractor target directory.

murdo-moj:
* the extractor data wasn't flowing to the new architecture in dev or preprod because the directory path to write to was incorrect. I am correcting this.


andrewrlee (approved):
LGTM

jakemulley (approved):

PR #13773, opened 2023-06-12
CFE-Civil: Add all Civil Apply namespaces to CFE-Civil

colinbruce:
Once the service is cutover we will rely on a stable environment This should allow all Civil Apply namespaces to call CFE-Civil production using internal routing for speed enhancements


davidread:
Closed in favour of https://github.com/ministryofjustice/cloud-platform-environments/pull/13780

PR #13887, opened 2023-06-14
TACT List - add cross-platform IAM role

vonbraunbates:
Allow pull bucket to get data from the Analytical Platform using an IRSA.
Also removes an unused RDS instance.


vonbraunbates:
Concourse reports:
`FATA[0014] error running terraform on namespace ct-tact-list-dev: unable to do Terraform Plan: variable "role_policy_arns" was required but not supplied `
I am using the suggestion [here](https://user-guide.cloud-platform.service.justice.gov.uk/documentation/other-topics/accessing-aws-apis-and-resources-from-your-namespace.html#using-irsa-in-your-namespace):
```
 # IRSA configuration
  service_account_name = "${var.team_name}-${var.environment}"
  namespace            = var.namespace # this is also used as a tag
  role_policy_arns = {
    s3 = module.s3.irsa_policy_arn
  }
```
whereas the template example [here](https://user-guide.cloud-platform.service.justice.gov.uk/documentation/other-topics/access-cross-aws-resources-irsa-eks.html) suggests that I should use the strings on the `data` line, e.g.:
```
module "irsa" {
      source = "github.com/ministryofjustice/cloud-platform-terraform-irsa?ref=1.0.2"
      namespace        = "my-namespace"
      role_policy_arns = [aws_iam_policy.my_namespace_my_policy.arn]
    }
data "aws_iam_policy_document" "my_namespace_my_policy" {
...
```
although that's from v1.0.2 of the IRSA template.

vonbraunbates:
After using [this irsa](https://github.com/ministryofjustice/cloud-platform-environments/blob/e7d80eeefcc8f20ad16eae776ead0cb27ed0f0bb/namespaces/live.cloud-platform.service.justice.gov.uk/hmpps-manage-adjudications-api-prod/resources/ap-data.tf) as an example, checks are now passing.

It looks like the problem was caused by `variables.tf` having lines for `role_policy_arns` and `service_account_name` when the former should be defined within the IRSA and the latter already has a default which doesn't need to be defined again.

jakemulley (approved):

vonbraunbates:
This will generate the role details for https://dsdmoj.atlassian.net/browse/JCH-294 .

PR #14414, opened 2023-07-07
PI-1226 Grant additional CI permissions (prod)

marcus-bcl:


namespaces/live.cloud-platform.service.justice.gov.uk/hmpps-probation-integration-services-prod/resources/variables.tf
```diff
@@ -86,10 +86,17 @@ variable "serviceaccount_rules" {
...
+    {
+      api_groups = [""]
+      resources  = ["pods/log"]
+      verbs      = ["get"]
+    },
     {
       api_groups = [""]
       resources = [
         "pods/portforward",
+        "pods/exec",
```
jaskaransarkaria, line 98: Just a note this allows the ci to exec into your application container, is this necessary?

marcus-bcl: Agree it's a bit dubious, and I'm definitely open to other suggestions.

The reason we're doing this is effectively to send messages to an SQS queue from a GitHub Actions workflow.  In the past we would have used AWS access keys, but with the move to IRSA we now need to run a pod in the namespace using the service account.  See thread: https://mojdt.slack.com/archives/C57UPMZLY/p1686298610170169?thread_ts=1686296526.348519&cid=C57UPMZLY

The script we run is here (it would be much cleaner if we didn't have to use the service pod approach): https://github.com/ministryofjustice/hmpps-probation-integration-scripts/blob/main/replay-messages.sh#L66-L88

jaskaransarkaria: It's not an ideal setup, but I'm happy to approve given that you have knowledge about it. Perhaps a better design would be a light API which runs as a pod and is protected by an API key, the API would receive a payload that t pushes onto the SQS

jaskaransarkaria (approved):

PR #14422, opened 2023-07-07
Changing CVL dev environment to use domain event

SebNegahbanMoJ:
Making an alteration to #14309 to use a domain event rather than a several prison events for updates to offender details


petergphillips (approved):
LGTM

poornima-krishnasamy (approved):

PR #14429, opened 2023-07-07
offender-management-production: Upgrade DB engine version to 14.7

ayqazi:
The actual DB engine version is 14.7 even though we configured 14.3 - update the TF to match reality and disable minor version upgrades in the future

### DIFF
--- a/namespaces/live.cloud-platform.service.justice.gov.uk/offender-management-production/resources/rds.tf
+++ b/namespaces/live.cloud-platform.service.justice.gov.uk/offender-management-production/resources/rds.tf
@@ -7,20 +7,21 @@
 module "allocation-rds" {
   source = "github.com/ministryofjustice/cloud-platform-terraform-rds-instance?ref=5.19.0"
 
-  vpc_name               = var.vpc_name
-  db_instance_class      = "db.m5.large"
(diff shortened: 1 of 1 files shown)


namespaces/live.cloud-platform.service.justice.gov.uk/offender-management-production/resources/rds.tf
```diff
@@ -17,7 +17,7 @@ module "allocation-rds" {
   environment-name       = "production"
   infrastructure-support = "manage-pom-cases@digital.justice.gov.uk"
   db_engine              = "postgres"
-  db_engine_version      = "14.3"
+  db_engine_version      = "14.7"
```
jaskaransarkaria, line 20: I would recommend explicitly setting this var to false

https://github.com/ministryofjustice/cloud-platform-terraform-rds-instance#input_allow_minor_version_upgrade

OR

if you are happy with minor upgrades then specify:

`db_engine_version = 14`

ayqazi: Ah, that's what went wrong.... OK will do that.

jaskaransarkaria (approved):

PR #15212, opened 2023-08-02
🔧  add the bucket name to the origin path for prisoner content hub dev cloudfront configuration.

Eli-TW:


Eli-TW:
After discussion with Jake, this is not the correct solution, so closing.

PR #15303, opened 2023-08-04
CDPS-320: Remove 05-serviceaccount-circleci.yaml and replace with ser…

davidmiddletonmoj:
…viceaccount.tf


namespaces/live.cloud-platform.service.justice.gov.uk/hmpps-digital-prison-services-prod/resources/serviceaccount.tf
```diff
@@ -0,0 +1,75 @@
+locals {
+  github_repos = ["hmpps-digital-prison-services"]
```
mattops, line 2: You dont need this line.

sj-williams (approved):

PR #15549, opened 2023-08-15
Civil Apply Production: Update long running prometheus rules

colinbruce:
We had shifted v1/bank_statements_controller into the 10 second window as file uploads regularly exceed 2 seconds and we, and users, are fine with it.  This is now being shown in the logs as v1/bank_statements so this updates the rule to prevent false positives in the alerts channel


jsugarman (approved):
looks good to me. I can see request URIs in kibana that would match. Makes me wonder, however, whether the exclusions like `providers/bank_statements_controller` would ever be effective - do you know what attribute/field of the logs is the `controller!~` checking for?

colinbruce:
> looks good to me. I can see request URIs in kibana that would match. Makes me wonder, however, whether the exclusions like `providers/bank_statements_controller` would ever be effective - do you know what attribute/field of the logs is the `controller!~` checking for?

Good shout, as discussed - I have updated both bank_statement endpoints to remove the `_controller` suffix

pwyborn (approved):

PR #15686, opened 2023-08-18
EL-1101: prevent e-mail spoofing (CCQ)

MazOneTwoOne:
Have added DNS records to prevent email spoofing of our production domain, following this [guidance](https://www.gov.uk/guidance/protect-domains-that-dont-send-email). Ideally would like a thorough review, as I may have missed something 

I could not create a valid null MX record.


jaskaransarkaria (approved):
LGTM

PR #15701, opened 2023-08-18
Adding debug user. Has to be preprod as no lower envs for legacy exist

carrb-moj:


carrb-moj:
Using irsa with a debug container, looks to give us what we need. Thanks

PR #16105, opened 2023-09-01
VB-2710 SQS non-association event configuration for DevOps, for PrePr…

AledWEvans:
VB-2710 SQS non-association event configuration for DevOps, for preprod


petergphillips (approved):
LGTM

PR #16239, opened 2023-09-05
RR-279 - Add service account to preprod namespace

mattwills-moj-digital:


mattwills-moj-digital:
Judging by other projects, it looks like this isn't required.

PR #16306, opened 2023-09-07
Share bucket arn across namespaces

StevenLeighton21:
The goal is to add access to the `json-output-s3-bucket` defined in f`ormbuilder-platform-test-dev` to the service account for the editor, `formbuilder-editor-test` which is defined in editor-service-account.yaml in the namespace `formbuilder-saas-test`.

Currently access is shared by exporting access key id and secret key in a kubernetes secret and injecting them via pipeline environment, this should replace that method by sharing a secret containing the arn to assume across namespaces and adding that iam role to the existing editor service account.


StevenLeighton21:
Not necessary - that bucket wasn't needed directly from the saas namespace

PR #16716, opened 2023-09-27
Increase redis instance size dev

itoulou:


jakemulley (approved):

itoulou:
Unnecessary to increase the redis cluster size

PR #17058, opened 2023-10-17
Create CircleCI service account in namespace: laa-govuk-notify-orchestrator-dev

BenMillar-CGI:
Created CircleCI service account


pwyborn (changes requested):
Hi  Ben @Ben Millar - sorry we have moved away from creating serviceaccounts using yaml as the secrets are not created automatically which is a deprecation from 1.24.
It creates problems in the pipeline run.
Please see here an example of service account utilising:
resources/serviceaccount-circleci.tf
https://github.com/ministryofjustice/cloud-platform-environments/commit/32a4c4ec3141b57355e8f5474791641ab6328b4c

PR #17193, opened 2023-10-23
SDIT-723: 🔧 Turn on auto tune for pre-prod

petergphillips:


namespaces/live.cloud-platform.service.justice.gov.uk/hmpps-prisoner-search-preprod/resources/opensearch.tf
```diff
@@ -14,12 +14,22 @@ module "hmpps_prisoner_search_opensearch" {
 
   cluster_config = {
     instance_count = 2
-    instance_type  = "t3.medium.search"
+    instance_type  = "m6g.large.search"
```
petergphillips, line 17: want to try out auto-tune in an environment other than production, so have to bump the instance here as `t3` doesn't support it

jackstockley89 (dismissed):

pwyborn (approved):

PR #17426, opened 2023-10-31
SDIT-670: 🚚 Move restricted patients queue to own namespace

petergphillips:


petergphillips:
turns out that we can't move queues, they have to be deleted and re-created

PR #17501, opened 2023-11-01
SDIT-670: 🚚 Remove irsa policy for old domain events queue for hmpps-domain-event-logger-prod

petergphillips:


prakash-moj (approved):
LGTM - Plan removes policy attachment as per the requirement.

PR #17645, opened 2023-11-08
Add lifecycle rule to laa-cla-backend-uat S3 Bucket

psweeting1:
Adds a lifecycle rule to the `laa-cla-backend-uat` S3 Bucket to expire entries that are over 2 years old.


said-moj:
@psweeting1 This change will remove the the file in the S3 bucket but what about the Import record which will be displayed as a link in the admins report dashboard

PR #17682, opened 2023-11-10
Temp solution - create alert for deschduling

vc13837:


poornima-krishnasamy:
@vc13837 , Is this PR still needed? Can you please have a look at the errors on the prometheus linter and amend it

PR #17915, opened 2023-11-24
MAP-465 move tag from bvl to domain events

GurnankCheema:
tag added to book-video-link-dev when should have added to domain-events-dev


GurnankCheema:
multi namespace PRs not allowed

PR #18079, opened 2023-12-01
IR-68: Fix #18074 – there must be at least 2 elasticache clusters

ushkarev:


namespaces/live.cloud-platform.service.justice.gov.uk/hmpps-incident-reporting-dev/resources/elasticache.tf
```diff
@@ -12,7 +12,7 @@ module "elasticache_redis" {
   node_type             = "cache.t4g.micro"
   engine_version        = "7.1"
   parameter_group_name  = "default.redis7"
-  number_cache_clusters = "1"
+  number_cache_clusters = "2"
```
ushkarev, line 15: was worth a try, eh? 1 would've been cheaper than 2 😄

jaskaransarkaria (approved):

PR #18225, opened 2023-12-11
Add nginx ingress deny to metrics endpoint for canary apps

mikebell:


mikebell:
Tested on live-2 and shows a 403 on /metrics endpoint as expected. Pain route and healthz endpoint unaffected.

jaskaransarkaria (approved):

PR #18237, opened 2023-12-12
request: dev namespace officehub-cms

ltd-pa:
Request to add a namespace for the dev of this AP app. 

@julialawrence any chance you could take a quick look? 

Many thanks!

### DIFF
--- a/namespaces/live.cloud-platform.service.justice.gov.uk/data-platform-app-officehub-cms-app-dev/01-rbac.yaml
+++ b/namespaces/live.cloud-platform.service.justice.gov.uk/data-platform-app-officehub-cms-app-dev/01-rbac.yaml
@@ -1,0 +1,16 @@
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: data-platform-app-officehub-cms-app-dev-admin
+  namespace: data-platform-app-officehub-cms-app-dev
+subjects:
+  - kind: Group
+    name: "github:officehub-cms-app-v2-admin-team"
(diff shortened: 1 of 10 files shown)


namespaces/live.cloud-platform.service.justice.gov.uk/data-platform-app-officehub-cms-app-dev/01-rbac.yaml
```diff
@@ -0,0 +1,16 @@
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: data-platform-app-officehub-cms-app-dev-admin
+  namespace: data-platform-app-officehub-cms-app-dev
+subjects:
+  - kind: Group
+    name: "github:officehub-cms-app-v2-admin-team"
```
julialawrence, line 8: This needs to be the team that's actually the admin of that repo. When we did migrations, we created the teams which is why they have such weird name but in your case, one already exists, so it should be added here.
In your case it should be `github:ltd-pa`

julialawrence (changes requested):
Can you please update the github name.

namespaces/live.cloud-platform.service.justice.gov.uk/data-platform-app-officehub-cms-app-dev/01-rbac.yaml
ltd-pa: Thanks - now resolved :)

prakash-moj (approved):

julialawrence (approved):

PR #18412, opened 2023-12-20
Add namespace laa-claim-non-standard-magistrate-fee-dev-crm4

jsugarman:
Add namespace laa-claim-non-standard-magistrate-fee-dev-crm4

### DIFF
--- a/namespaces/live.cloud-platform.service.justice.gov.uk/laa-claim-non-standard-magistrate-fee-dev-crm4/00-namespace.yaml
+++ b/namespaces/live.cloud-platform.service.justice.gov.uk/laa-claim-non-standard-magistrate-fee-dev-crm4/00-namespace.yaml
@@ -1,0 +1,16 @@
+apiVersion: v1
+kind: Namespace
+metadata:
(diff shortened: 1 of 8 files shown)


.checksum
```diff
@@ -1,3 +1,3 @@
 #This file is used by the auto pr github action. Please commit
-laa-crime-equinity-historical-data-uat
-h1:RuS6kR5y1nTdzyxP79Wh4YESo2e0iqBr+uNcew3VRug=
+laa-claim-non-standard-magistrate-fee-dev-crmfour
```
jsugarman, line 2: I assume this "old" name in the checksum is irrelevant?

I had to name it `..-crmfour` because the cli does not allow numbers in namespace name, github team and email addresses.

poornima-krishnasamy: Hi @jsugarman, the checksum hash is generated from the namespace name. So it will be a problem for next namespace generated from cli. can you skip that file from your PR commit.

poornima-krishnasamy (approved):

PR #18446, opened 2023-12-21
CMR547-854: correct namespace for shared secret)

jsugarman:
CMR547-854: correct namespace for shared secret

This secret needs to generated in a different namespace
that requires access to the same s3 bucket.


jsugarman:
> #### Namespace Mismatch Found:
> Show
> *Pusher: @jsugarman, Workflow: `Compare Respository Namespace with Terraform Resource Namespace`*

This is deliberate

sj-williams (approved):

PR #18985, opened 2024-01-22
HAAR-2243: set prepare_for_major_upgrade to false and

SimonMitchellMOJ:
remove skip namespace file for hmpps-authorization-server-dev


tmahmood72 (approved):
LGTM

PR #19000, opened 2024-01-23
Add Omagh inquiry domain MX mail records

brown-a2:
New MX records for the omagh.independent-inquiry.uk domain.


tmahmood72 (approved):
LGTM

PR #19101, opened 2024-01-25
Remove POC Namespace not in use

pking-moj:
Namespace was created to PoC a concept within the team. Tearing down now we have what we need and is not in use.


tmahmood72 (approved):
LGTM

PR #19188, opened 2024-01-29
request for a namespace to be created so i can push a test app

FolarinOyenuga:
Hi,

I wish to have the created namespace validated so a test app could be subsequently pushed. Thanks.


tmahmood72 (commented):
Hi Folarin,

1) Please can you update the team name to:
name: "github:webops" in file 01-rbac.yaml

2) The namespace should be in a format like folarin-test or folarein-dev

3) the folder name should be the same as the new name (above) in file 00-namespace.yaml

PR #19207, opened 2024-01-29
Upgrade rds minor version for laa-court-data-ui [PROD]

Obsiye:
Upgrade rds minor version for `laa-court-data-ui-prod`

- Update the postgresql minor version to 14.10 from 14.7

This is to mitigate many CVEs. Also, 14.7 will soon no longer be supported.


tmahmood72 (approved):
LGTM

PR #19209, opened 2024-01-29
EMP-3: Update MSSQL engine used in PROD RDS

tanovellino:
Update MSSQL engine used in PROD RDS to -web, was -ex
Also, an additional increase in size now that we are starting the real data migration


tmahmood72 (approved):
LGTM

PR #19238, opened 2024-01-30
For DCES-248. Removing the API Gateway as this is not required. The r…

atharmajeed:
For DCES-248. Removing the API Gateway as this is not required. The requests will be handled by Ingress Controller.


tmahmood72 (approved):
LGTM

PR #19271, opened 2024-01-31
Rename `rds` `snapshot_identifier` and secret name for `laa-crown-court-remuneration-dev`

mpw5:
The current values are causing issues as they conflict with the same config in `laa-crown-court-remuneration-sandbox-admin`


tmahmood72 (approved):
LGTM

PR #20278, opened 2024-02-02
PI-1841 apply db upgrade 

anthony-britton-moj:
Previous attempt to upgrade to postgres 16 failed because version wasn't at 14.10 beforehand, so minor update to 14.10 first to permit major upgrade to 16 afterwards


tmahmood72 (approved):
LGTM

PR #20517, opened 2024-02-05
DSOS-2543: add oidc role for offlice file transfer

jnq:
use the oidc auth to assume the role in GHA, negates the need for token sharing


petergphillips (dismissed):
LGTM

PR #20836, opened 2024-02-13
MTP-1980: Set `prepare_for_major_upgrade` to `false`

xoen:
and revert DB version.


xoen:
Hi @timckt - do you think this should put the terraform state in sync with what's in AWS and allow me to upgrade the prod DB to `16.1` in a separate PR?

PR #21167, opened 2024-02-23
Update 04-networkpolicy.yaml - Cymulate IP Range for Pen Testing

neiltait:
**What this change is:**

We have added the [Cymulate](https://cymulate.com/) egress IP range supplied by MoJ internal pen-testers to the inbound whitelist.

We have also added the IP of one of our dev laptops so we can verify the change prior to pen-testing.

**Why this change is required**

The MoJ internal pen-test uses a cloud-based tool, Cymulate, and this needs to access our build in Dev to run the tests.

**NOTE**

This is temporary and will be removed following the pen-test.


neiltait:
Closed as we need to do this differently

PR #21242, opened 2024-02-26
Chore/update cas test rds

tahb:
- We were asking to update the CAS postgres minor version due to it soon being deprecated. 
- We found that it has already been moved from 14.7 to 14.10
- While we're here we want to make this more automated going forwards so we enable automatic minor updates. Before we do we assign an explicit maintenance window so we are clear about avoiding disruption to users
- While we're here we also make test no longer auto update _major_ versions so it behaves like prod, providing greater environment parity

### DIFF
--- a/namespaces/live.cloud-platform.service.justice.gov.uk/hmpps-community-accommodation-test/resources/rds.tf
+++ b/namespaces/live.cloud-platform.service.justice.gov.uk/hmpps-community-accommodation-test/resources/rds.tf
@@ -13,6 +13,8 @@ module "rds" {
   db_instance_class            = "db.t3.small"
   rds_family                   = "postgres14"
   allow_major_version_upgrade  = "false"
+  backup_window                = var.rds_backup_window
+  maintenance_window           = var.rds_maintenance_window
 
   providers = {
     aws = aws.london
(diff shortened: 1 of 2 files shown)


namespaces/live.cloud-platform.service.justice.gov.uk/hmpps-community-accommodation-test/resources/rds.tf
```diff
@@ -12,7 +12,11 @@ module "rds" {
   db_engine_version            = "14"
   db_instance_class            = "db.t3.small"
   rds_family                   = "postgres14"
-  allow_major_version_upgrade  = "true"
+  allow_major_version_upgrade  = "false"
+  allow_minor_version_upgrade  = "true"
+
+  backup_window                = var.rds_backup_window
```
tahb, line 18: I borrowed the configuration of these from [Locations Inside Prison](https://github.com/ministryofjustice/cloud-platform-environments/blob/8da1196b0ccec1912a8877b8f8e2fc2177aec759/namespaces/live.cloud-platform.service.justice.gov.uk/hmpps-locations-inside-prison-prod/resources/rds.tf#L19).

FolarinOyenuga (changes requested):
Your commit has some conflicts.

PR #21244, opened 2024-02-26
Chore/update cas dev rds

tahb:
- We were asking to update the CAS postgres minor version due to it soon being deprecated.
- We found that it has already been moved from 14.7 to 14.10
- While we're here we want to make this more automated going forwards so we enable automatic minor updates. Before we do we assign an explicit maintenance window so we are clear about avoiding disruption to users
- While we're here we also make test no longer auto update major versions so it behaves like prod, providing greater environment parity


FolarinOyenuga (approved):

FolarinOyenuga (changes requested):
Could you resolved the conflicts?

PR #21268, opened 2024-02-27
VB-3550 Put application refactor live, had a db size issue when putti…

AledWEvans:
increased the size of DB as 

SQL State : 53100 
ERROR: could not write to file "base/pgsql_tmp/pgsql_tmp468.0": No space left on device

Error on the DB while migrating data, this should hopefully solve the problem


tmahmood72 (approved):
LGTM

PR #21271, opened 2024-02-27
HAAR-2363: 🔧 Rename service account, role and rolebinding for hmpps-external-users-api-dev

SimonMitchellMOJ:


tmahmood72 (approved):
LGTM

PR #21614, opened 2024-03-06
CFE-CIVIL PRODUCTION add skip file for namespace

willc-work:
We are upgrading the DB for CFE Civil Production tomorrow morning. In preparation this PR adds the skip file needed for when we merge the DB PR.


tmahmood72 (approved):
LGTM

PR #21623, opened 2024-03-06
CFE Civil PROD update pingdom url

willc-work:
updating the URL pingdom uses to check the service is up.

the new status endpoint has a DB check

QUESTION:
We have the APPLY_SKIP_PIPELINE file on the namespace for a DB upgrade. Will that affect this PR ie. will it prevent this change from occurring until after that skip file is removed?


davidread:
>  We have the APPLY_SKIP_PIPELINE file on the namespace for a DB upgrade. Will that affect this PR ie. will it prevent this change from occurring until after that skip file is removed?

I believe so. In the meantime you could try editing the value in pingdom manually.

tmahmood72 (approved):
LGTM

PR #21626, opened 2024-03-06
Change to actual current version of DB - peoplefinder-production

vertism:
Apply is erroring, so changing to current version.


tmahmood72 (approved):
LGTM

PR #21988, opened 2024-03-19
validate annotations for namespaces from PRs targeting the main branch

FolarinOyenuga:
This GitHub action workflow automatically checks, along with the validate-annotations.go script, if:

1. cloud-platform.justice.gov.uk/source-code
    cloud-platform.justice.gov.uk/team-name
    
    are valid entries.

2. It also contains unit tests for the validation script.


cmd/annotations-checker/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM golang:1.19-alpine
```
jaskaransarkaria, line 1: Can you switch to `debian:bookworm-slim` please

cmd/annotations-checker/validate-annotations.go
```diff
@@ -0,0 +1,119 @@
...
+	"path/filepath"
+
+	"github.com/google/go-github/v32/github" // GitHub API client library
+	// "github.com/ministryofjustice/cloud-platform-environments/pkg/namespace"
+	"golang.org/x/oauth2" // API requests authentication
+	// "google.golang.org/genproto/googleapis/api/annotations"
+	"gopkg.in/yaml.v2"
+)
+
+// Annotations structs to define the structure of annotations in the YAML file
```
jaskaransarkaria, line 16: Let's get rid of all these comments, if you find a need for a comment try renaming the thing that needs a comment

.github/workflows/validate-annotations.yml
```diff
@@ -0,0 +1,50 @@
...
+    - name: Set up Go
+      uses: actions/setup-go@v3
+      with:
+        go-version: 1.19
+      
+      # execute the validation script
+    - name: Check out code
+      uses: actions/checkout@v3
+      
+    - name: Install dependencies
```
jaskaransarkaria, line 31: We don't need this step if we are running in a docker container

.github/workflows/validate-annotations.yml
```diff
@@ -0,0 +1,50 @@
...
+
+    - name: Build Docker image
+      run: docker build -t validate-annotations .
+
+    - name: Run Validation
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      run: docker run --rm -e GITHUB_TOKEN=$GITHUB_TOKEN validate_annotations
+
+    - name: Run Validation (Local)
```
jaskaransarkaria, line 46: We don't need this step if we are running in a docker container

.github/workflows/validate-annotations.yml
```diff
@@ -0,0 +1,50 @@
...
+      
+      # execute the validation script
+    - name: Check out code
+      uses: actions/checkout@v3
+      
+    - name: Install dependencies
+      run: |
+        go mod download
+
+    - name: Build
```
jaskaransarkaria, line 35: We don't need this step if we are running in a docker container

cmd/annotations-checker/validate-annotations.go
```diff
@@ -0,0 +1,119 @@
...
+		Annotations Annotations `yaml:"annotations"`
+	} `yaml:"metadata"`
+}
+
+// gitHub client
+func NewGitHubClient(token string) *github.Client {
+	ctx := context.Background()
+	ts := oauth2.StaticTokenSource(&oauth2.Token{AccessToken: token})
+	tc := oauth2.NewClient(ctx, ts)
+	client := github.NewClient(tc)
```
jaskaransarkaria, line 35:
```suggestion
	return github.NewClient(tc)
```

cmd/annotations-checker/validate-annotations.go
```diff
@@ -0,0 +1,119 @@
...
+	return client
+}
+
+// validate team name and source code repo
+func ValidateAnnotations(client *github.Client, org string) (string, error) {
+	// get namespace directory
+	namespacesDir := "/path/to/namespaces"
```
jaskaransarkaria, line 42: update this, you need to find a way of passing the files that have been changed/ added and the namespace(s) that are mentioned from the PR to the code

cmd/annotations-checker/validate-annotations.go
```diff
@@ -0,0 +1,119 @@
...
+	files, err := os.ReadDir(namespacesDir)
+	if err != nil {
+		return "", fmt.Errorf("Failed to read namespaces directory: %v", err)
+	}
+
+	// check if kind:Namespace exists in any file
+	foundNamespace := false
+	for _, file := range files {
```
jaskaransarkaria, line 50: Currently, this will loop over every file in the namespace (so it will always trigger). We only want to check new files or files that have been modified. We don't care about files that already exist (we can address those later) and we don't care about files that are being deleted either.

If you struggle to figure out a way to do this for your current code consider calling a git diff on the PR against main or exploring passing down the `github.context` (specifically accessing the event) from the action [this](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context) and [this](https://stackoverflow.com/a/77153325) link might be helpful.

.github/workflows/validate-annotations.yml
```diff
@@ -0,0 +1,50 @@
+# this action is meant to trigger the workflow on pull requests aimed at the main branch that include changes in the namespaces/ directory
```
jaskaransarkaria, line 1: unnecessary comment

.github/workflows/validate-annotations.yml
```diff
@@ -0,0 +1,50 @@
...
+name: Validate Annotations
+
+on:
+  pull_request:
+    branches:
+      - main
+    paths:
+      - 'namespaces/**'
+
+
```
jaskaransarkaria, line 12: unnecessary newline

.github/workflows/validate-annotations.yml
```diff
@@ -0,0 +1,50 @@
...
+jobs:
+
+  validate-annotations:
+    # env where the job will be executed
+    name: Validate Annotations
+    runs-on: ubuntu-latest
+    steps:
+
+      #set up a Go env
+    - name: Set up Go
```
jaskaransarkaria, line 22: We don't need this step if we are running in a docker container

.github/workflows/validate-annotations.yml
```diff
@@ -0,0 +1,50 @@
...
+      uses: actions/setup-go@v3
+      with:
+        go-version: 1.19
+      
+      # execute the validation script
```
jaskaransarkaria, line 27: unnecessary comment

.github/workflows/validate-annotations.yml
FolarinOyenuga: The whole steps or Run Validation (Local)?

PR #22427, opened 2024-04-15
removed pipeline skip

danielglen-moj:
As part of the DR test we carried out we needed to add a pipeline skip. This PR removes this due to DR test finishing


tmahmood72 (approved):
LGTM

PR #22460, opened 2024-04-15
CRS-1883 add an alternative dns

joelstobart-moj:
Add DNS for the alternative CRDS system (to run in a different configuration) in preprod frontend


tmahmood72 (approved):
LGTM

PR #22463, opened 2024-04-16
Creating a new namespace for a dev environment for opg-actor-search app

khb28:
Creating a namespace as a step to deploy a new app


khb28:
I'm not sure what the check is that hasn't completed. If it's an issue please can you advise.

PR #22515, opened 2024-04-17
CPR-276 create machine readable S3 bucket for nDelius extract

markreesmoj:
The intention is to create a bucket which is only readable by the IRSA account named person-record-service - i.e. it is not readable by any human user.


namespaces/live.cloud-platform.service.justice.gov.uk/hmpps-person-record-prod/resources/s3-ndelius-extract.tf
```diff
@@ -0,0 +1,76 @@
...
+                "StringEquals": {
+                    "s3:x-amz-acl": "bucket-owner-full-control"
+                }
+            }
+        },
+        {
+             "Sid": "S3ListPolicy",
+             "Effect": "Allow",
+             "Principal": {
+                 "AWS": "arn:aws:iam::050243167760:root"
```
jaskaransarkaria, line 47: if "The intention is to create a bucket which is only readable by the IRSA account named person-record-service" then why not use that role arn here?

jaskaransarkaria (approved):

PR #22556, opened 2024-04-18
VB-3379, remove pipeline skip file, set major_upgrade to false

hutcheonb-moj:
Remove pipeline skip file
Set `major_upgrade` back to false

All upgrades completed, this is just the cleanup PR


tmahmood72 (approved):
LGTM

PR #22562, opened 2024-04-18
CRM457-1288: Production Provider Network Policy

ivanELEC:
Tickets
[Update network policies so that Provider application talks to datastore via Cloud Platform rather than via an external url](https://dsdmoj.atlassian.net/browse/CRM457-1288)


tmahmood72 (approved):
LGTM

PR #22607, opened 2024-04-22
Update variables.tf

khb28:
Attempting to provide a kubernetes_cluster


khb28:
test didn't work - don't want to make the change after all

PR #22620, opened 2024-04-23
BAVL-13 add missing rbac group required by the HAAR team to add auth kube secrets.

matt-crowson-moj:
This group is required by the HAAR team so they can add the secrets required for HMPPS Auth.


tmahmood72 (approved):
LGTM

PR #22659, opened 2024-04-24
Adding S3 DeleteObject permission to IAM policy

carrb-moj:
Had PutObject twice within IAM policy statement and missed out DeleteObject.


tmahmood72 (approved):
LGTM

PR #23341, opened 2024-05-30
LCAM-1325|Update network policies to allow service-to-service communi…

venkat980:
…cation with other microservices


tmahmood72 (approved):
LGTM

PR #23548, opened 2024-06-12
Bumps ecr in staging to latest module version

ross75:


ross75:
Wrong env - needs dev.

PR #23569, opened 2024-06-13
Add Environment Lookup

harichintala1:


namespaces/live.cloud-platform.service.justice.gov.uk/hmpps-digital-prison-reporting-mi-dev/resources/cross-iam-role-irsa.tf
```diff
@@ -20,7 +28,7 @@ module "dpr_mi_assume_role" {
   role_name                     = "dpr-reporting-mi-${var.environment}-cross-iam-${var.eks_cluster_name}"
   provider_url                  = data.aws_eks_cluster.eks_cluster.identity[0].oidc[0].issuer
   role_policy_arns              = [aws_iam_policy.cross_iam_dpr_oidc.arn]
-  oidc_fully_qualified_subjects = ["system:serviceaccount:${var.namespace}:dpr-reporting-mi-development-cross-iam"]
+  oidc_fully_qualified_subjects = ["system:serviceaccount:${var.namespace}:dpr-reporting-mi-${lookup(local.environments_map, lower(var.environment))}-cross-iam"]
```
gavriil-g-moj, line 31: This is in the hmpps-digital-prison-reporting-mi-dev folder.
Is it OK to be there or would it be better if we had separate cross-iam-role-irsa.tf files for each of the environment folders?




[9:06](https://mojdt.slack.com/archives/D05C1J9SN5R/p1718266008880379)
Is it OK to be there or would it be better if we had separate cross-iam-role-irsa.tf files for each of the environment folders?

PR #23623, opened 2024-06-14
SAA-1785 - Enable data extractor to pull data from activities dev env for AP

KristianGregor:
Configures the data extractor for extracting data to AP from the activities management (Dev) service

Configuration as per guidance:
https://user-guide.cloud-platform.service.justice.gov.uk/documentation/other-topics/getting-your-data-onto-the-analytical-platform.html#how-does-data-from-my-service-get-onto-the-analytical-platform


KristianGregor:
Will take a different approach

PR #23809, opened 2024-06-24
chore: 🤖 bump malformed-yaml action

sj-williams:
update to latest malformed-yaml release with fix for multi doc yaml, plus test a multi doc yaml


tmahmood72 (approved):
LGTM

PR #23853, opened 2024-06-25
Appstore PROD: Add smon and netpol permissions

jsugarman:
- **Remove hpa and prometheusrules permissions**
- **Add smon and netpol**


tmahmood72 (approved):
LGTM

PR #23862, opened 2024-06-25
BAVL-213 adding subscription to the video booking cancelled domain event for the BVLS service.

matt-crowson-moj:
Adding subscription to a new BLVS domain event to for the BVLS service for processing the cancellation of booking events.


tmahmood72 (approved):
LGTM

PR #23871, opened 2024-06-26
Remove incorrect namespace

titlescreen:
This was made with the name pasted in twice.
I cannot rename it so I need to delete this and
then make a new one.


tmahmood72 (approved):
LGTM

PR #24086, opened 2024-07-03
Delete service pod sys int api dev

popey2700:
Deleting no longer required service pod


FolarinOyenuga:
Hi @popey2700 is this request still needed? If no, would you mind closing, please?

PR #24177, opened 2024-07-08
Upgrade rds from xlarge to 2xlarge

ArronChave3:
Upgrade the RDS instance from xlarge to 2xlarge


FolarinOyenuga:
Hi @ArronChave3 is this request still needed? If no, would you mind closing, please?

PR #24326, opened 2024-07-11
Saa 1785 activities upload to ap s3

KristianGregor:
Creates the service account that will be used by Activities and Appts service to upload data to Analytical Platform S3 buckets part of Register-My-Data

This is DEV only


KristianGregor:
This needs some changes, further work needed

FolarinOyenuga (approved):

PR #24543, opened 2024-07-17
Update github team for CCLF dev console access

mpw5:
The GithubTeam value in `default_tags` in `main.tf` is set to the team previously responsible for this application. This updates the value to match the team defined in `01-rbac.yaml` (`laa-lz-cp-migration`) so that all relevant developers can access the console for this application.


tmahmood72 (approved):
LGTM

PR #24587, opened 2024-07-18
Add missing PrometheusRule for datahub prod

MatMoore:
Part of https://github.com/ministryofjustice/data-catalogue/issues/187

The rule for production is copy pasted from preprod, but with the URLs changed for RDS, OpenSearch and the grafana dashboard.


tmahmood72 (approved):
LGTM

PR #24702, opened 2024-07-24
VB-3901 - Remove old change event, ready for replacement with upserted and deleted events

adaviesMOJ:


adaviesMOJ:
Will re-open 1 per env.

PR #24815, opened 2024-07-26
[SI-1091] Subscribe cat tool to prisoner released events

a6software:
Hopefully this is correct, but is something of a guess as to how this process works.


jackstockley89 (approved):

Stubbs (commented):
I'm not sure that the `prisoner-offender-search.prisoner.released` events come from the  `offender_events` topic, I think you need another block to watch the `hmpps-domain-events-topic-arn` like this change below:

https://github.com/ministryofjustice/cloud-platform-environments/blob/cfa13269263a6d4dd26d05d0017e307c66fc17cc/namespaces/live.cloud-platform.service.justice.gov.uk/offender-events-dev/resources/pathfinder-sub-queue.tf#L209

a6software:
hi @jackstockley89 - would you be able to confirm whether or not I need to add the extra config Stuart suggests? https://github.com/ministryofjustice/cloud-platform-environments/pull/24815#pullrequestreview-2201549411

PR #24905, opened 2024-07-30
fix dns name issue for ui project for contacts

prabash-moj:
fix template issue with certificate


namespaces/live.cloud-platform.service.justice.gov.uk/hmpps-contacts-dev/06-certificate.yaml
```diff
@@ -23,4 +23,4 @@ spec:
     name: letsencrypt-production
     kind: ClusterIssuer
   dnsNames:
-    - contacts-dev.hmpps.service.justice.gov.uk
+    - contacts-ui-dev.hmpps.service.justice.gov.uk
```
timharrison-moj, line 26: We don't need this one, Prabash. This would undo the change in the other PR and make the cert contain the name contacts-ui-dev.xx.xx.xx. when we want contacts-dev.xx.xx.   Just close this PR, and let the other go through, and it will be all good.  Thanks.

PR #24914, opened 2024-07-31
add comment to hmpps-template-typescript

james-jdgtl:
Minor change to request that when the hmpps-template-typescript namespace entry is used as a template, not to remove the two groups that are in the rbac file.


tmahmood72 (approved):
LGTM

PR #24932, opened 2024-08-01
Removed route53 zone from dev

murdo-moj:
* According to CP guidance, only one hosted zone is needed for our project, and environments can use this zone
* ExternalDNS should create the DNS records for this and delegate eg dev requests to dev from the prod hosted zone, which exists
* Currently `dig NS dev.find-moj-data.service.justice.gov.uk +short` returns no nameservers
* I am hoping deleting this dev zone will trigger externalDNS to sort the domains and the nameservers should be returned from dev


FolarinOyenuga (approved):

FolarinOyenuga (changes requested):
Before proceeding with this, could you run "dig NS dev.find-moj-data.service.justice.gov.uk +trace" to see the entire DNS chain? Could point at the root cause.

murdo-moj:
I get a timeout running that command:
`;; connection timed out; n`

FolarinOyenuga (approved):

PR #25007, opened 2024-08-06
Make resources visible to secure-estate-digital-team

atulloch:


atulloch:
Should allow inspecting of the S3 bucket via AWS console.

PR #25028, opened 2024-08-06
Increase number of clauses allowed in query

wilson1000:
This update attempts to fix an error on the application where the query is carrying too many search clauses.


tmahmood72 (approved):
LGTM

PR #25156, opened 2024-08-13
EMP-439: Adding ServiceMonitor and NetworkPolicy for Prometheus in PROD

tanovellino:
EMP-439: Adding ServiceMonitor and NetworkPolicy for Prometheus in PROD


tmahmood72 (approved):
LGTM

PR #25164, opened 2024-08-13
do not turn of prod rds overnight

LavMatt:
We don't want our prod backend to shutdown overnight as we'll be running ingestions


tmahmood72 (approved):
LGTM

PR #25169, opened 2024-08-13
Give the client admin team permissions in hmpps-ems-cemo-ui-prod

devarrowsmith:
Grant the client admin team permission to access secrets in namespace hmpps-ems-cemo-ui-prod.


tmahmood72 (approved):
LGTM

PR #25179, opened 2024-08-14
Production resource requirements

wilson1000:


namespaces/live.cloud-platform.service.justice.gov.uk/intranet-production/resources/service-account.tf
EarthlingDavey, line 1: Was deleting this intentional?

namespaces/live.cloud-platform.service.justice.gov.uk/intranet-production/02-limit-range.yaml
```diff
@@ -6,9 +6,9 @@ metadata:
...
   limits:
     - default:
-        cpu: 1000m
-        memory: 1000Mi
+        cpu: 2000m
+        memory: 2000Mi
       defaultRequest:
-        cpu: 10m
-        memory: 100Mi
+        cpu: 500m
```
EarthlingDavey, line 12: The defaultRequests here are used by hpa.

Based on this in hpa.yml

```
  metrics:
    - type: Resource
      resource:
        name: cpu
        target:
          type: Utilization
          averageUtilization: 60
```

and, in staging, having `defaultRequest.cpu: 200m`.

My understanding is that we will scale up when cpu usage hits 60% of 200m - that's 120 milli cores.

If on production we have `defaultRequest.cpu: 500m` then we are changing the threshold for scaling to 60% of 500m - 300 milli cores.

That difference might mean that our auto scaling behaves differently to what we've seen in our stress tests.

EarthlingDavey (dismissed):

tom-j-smith (approved):

PR #25651, opened 2024-09-02
SI-658 Updated queue sub config to enable released/received events on prod.

Stubbs:


jonhindle (approved):
Looks good to me 👍
I'm the lead developer for the pathfinder service.

sj-williams (approved):

PR #25756, opened 2024-09-05
Make custom metrics available for HPA.

EarthlingDavey:
## Background

Our php-fpm application does not horizontally scale well based on CPU or memory metrics.

Our pods expose custom metrics, that are more suitable for HPA, and they are ingested by prometheus.

One metric in particular `phpfpm_active_processes` looks to be a good metric for pod pressure and an indicator for HPA.

This article explains the process that I'm following

https://aspirin4k.medium.com/scaling-of-php-application-in-kubernetes-based-on-fpm-workers-utilization-15e5c7d8683a

## This PR

To scale on a custom metric, it looks like we need to add an `adapter-config`.

The added config lets us scale on `phpfpm_active_processes` by exposing the `phpfpm_active_processes_utilization` metric to use in our HPA config.

### DIFF
--- a/namespaces/live.cloud-platform.service.justice.gov.uk/monitoring/adapter-config-php-fpm.yaml
+++ b/namespaces/live.cloud-platform.service.justice.gov.uk/monitoring/adapter-config-php-fpm.yaml
@@ -1,0 +1,19 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: adapter-config-php-fpm
+  namespace: monitoring
+data:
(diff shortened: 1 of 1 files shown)


EarthlingDavey:
Having read [this guide](https://gcore.com/learning/prometheus-metrics-based-autoscaling-in-kubernetes/), I think I'm missing the part where the config is added to the prometheus deployment.

Would it be possible for the team to look into this and advise?

Many thanks.

tmahmood72 (approved):
LGTM

PR #25757, opened 2024-09-05
Make Find MoJ data dev alerts more forgiving (dev)

MatMoore:
1. bump up the threshold for latency alerts
2. replace CPU alert with an autoscaler alert


murdo-moj (approved):
Looks good to me. The latency one might need a bit more tweaking/further investigation like you say but this is fine for now.

tmahmood72 (approved):
LGTM

PR #25775, opened 2024-09-06
PIC-4134 create new IRSA with SNS and S3 permissions

markreesmoj:
This will be used by court-hearing-event-receiver and crime-portal-gateway
Future PRs will remove these S3 and SNS permissions from the court-case-service IRSA


jaskaransarkaria (approved):

jaskaransarkaria (changes requested):
Error: Reference to undeclared module


  on service-pod.tf line 5, in module "service_pod":

   5:   service_account_name = module.irsa.service_account.name

markreesmoj:
> Error: Reference to undeclared module
> 
> on service-pod.tf line 5, in module "service_pod":
> 
> 5: service_account_name = module.irsa.service_account.name

Thanks, interesting.

Our plan is to have two service accounts with different permissions. Would we have to have two service pods in that case or is there a way to have a pod with both service accounts?

jaskaransarkaria (dismissed):

jaskaransarkaria (approved):

PR #25877, opened 2024-09-12
PIC-4134 add new service account with access to court-cases topic

markreesmoj:
This should not modify the existing irsa (to avoid disruption to live) and just add a new one

It would be great if you could add a release to https://github.com/ministryofjustice/cloud-platform-terraform-service-pod which currently shows 1.0.1 as the latest version


tmahmood72 (approved):
LGTM

sj-williams (approved):

PR #26186, opened 2024-09-25
TM-517, adding LAA production VPC CIDR to CCR Staging and UAT RDS

mnasr-moj:
Adding LAA Production VPC CIDR to CCR Staging and UAT RDS so that the windows EC2 instance I created in LAA Production can be used to access the DBs in CP. https://dsdmoj.atlassian.net/browse/TM-517


mnasr-moj (commented):
Made the change

PR #26264, opened 2024-09-26
chore: test rds io2 IOPS Not Specified

timckt:
Scenario 5: io2 IOPS Not Specified

- Variables:
  - storage_type = "io2"
  - db_iops = null

Result: Precondition fails with the error message about io2 IOPS.


timckt:
```
FATA[0028] error running terraform on namespace tim-development: unable to do Terraform Plan: exit status 1

Error: Resource precondition failed

  on .terraform/modules/rds/main.tf line 219, in resource "aws_db_instance" "rds":
 219:       condition     = var.storage_type != "io2" || (var.db_iops != null ? var.db_iops >= 1000 : false)
    ├────────────────
    │ var.db_iops is null
    │ var.storage_type is "io2"

When 'storage_type' is 'io2', 'db_iops' must be specified and at least 1000.
 

```

PR #26389, opened 2024-09-30
Remove messaging queue config

Ross-Nation:
The following is now redundant so we want to remove the config: 
laa_status_update_queue
laa_status_update_queue_policy
laa_status_update_dead_letter_queue


tmahmood72 (approved):
LGTM

PR #26417, opened 2024-09-30
Exclude infinities from latency alert expression

MatMoore:
We were computing an average by dividing the sum by the count. The count can be zero.

If the sum is 0, the expression returns NaN, and the comparison returns false.

But if the sum is non-zero, it gives +Inf, and we get an alert.

I'm not sure why we would have data points with 0 for the count, and non-zero for the sum, but we are definitely getting +Inf values and unwanted alerts, so I'm explicitly filtering out +Inf (if anyone knows a better way of fixing this, please let me know!)


murdo-moj (approved):

tmahmood72 (approved):
LGTM

PR #26633, opened 2024-10-07
Create environment cfe-civil-cicd-poc-uat

davidread:
It is a copy of cfe-civil-uat, but renaming the app from cfe-civil to cfe-civil-cicd-poc, which corresponds to the repo name being https://github.com/ministryofjustice/cfe-civil-cicd-poc


tmahmood72 (approved):
LGTM

PR #26635, opened 2024-10-07
⬆️ Bump ECR Module to 7.1.0

jasonBirchall:
This release is currently in pre-release. It creates the registry as a secret, meaning we don't expose the account number in plain text.


tmahmood72 (approved):
LGTM

PR #26749, opened 2024-10-10
[SI-1255] Upgrade cat tool pp cloud-platform-terraform-rds-instance to v7.3.1

a6software:
Following changes made during Cat Tool prod outage (SI-1254) it was flagged (https://mojdt.slack.com/archives/C04JFG3QJE6/p1728319701004019?thread_ts=1728311264.704499&cid=C04JFG3QJE6) that we could upgrade our rds instance to 7.3.0 to take advantage of faster disk resources.


a6software:
Bump to 7.3.1 based on build failure

PR #26852, opened 2024-10-14
Increase CCLF production prometheus threshold

mpw5:
The current threshold leads to noisy alerts. Increase the threshold to a more accurate value for this environment.


tmahmood72 (approved):
LGTM

PR #26961, opened 2024-10-18
Add hmpps-sre to hmpps-audit-preprod

james-jdgtl:
Required to be able to support when OpenSearch is implemented


james-jdgtl:
Restricted namespace - this will not be added.

PR #26964, opened 2024-10-18
Add hmpps-sre to RBAC for calculate-journey-variable-payments-prod

james-jdgtl:
This is required for when OpenSearch is implemented, to assist with support


tmahmood72 (approved):
LGTM

PR #27011, opened 2024-10-21
Replace dps-tech in create-and-vary-a-licence-api-preprod with hmpps-sre

james-jdgtl:
Part of HEAT-371 to deprecate dps-tech


tmahmood72 (approved):
LGTM

PR #27131, opened 2024-10-24
MAP-1724 update RDS module to 8.0.0

GurnankCheema:


GurnankCheema:
applied to dev instead of preprod

PR #27270, opened 2024-10-29
Portfolio management ssm set

james-jdgtl:
Create a set of ssm entries for application insight connection strings


sj-williams (approved):

james-jdgtl (commented):
done

james-jdgtl:
I've reverted the change to the other namespace.. hopefully this will go through now..

PR #27427, opened 2024-11-01
Upgrade RDS TF module to v8 for IRS (`dev`)

xoen:
This will change the storage type from `gp2` to `gp3`. The module also bumps the default storage size to 20GB (we don't specify this and it was previously
default to 10GB AFAICT).


tmahmood72 (commented):
LGTM

xoen:
@tmahmood72 thanks for the review. Not sure why but GH still says:

> **Merging is blocked**
> Merging can be performed automatically with 1 approving review.

and doesn't let me merge.

sj-williams (approved):

PR #27429, opened 2024-11-01
Upgrade RDS TF module to v8 for INC (`dev`)

xoen:
This will change the storage type from `gp2` to `gp3`. The module also bumps the default storage size to 20GB (we don't specify this and it was previously default to 10GB AFAICT).


tmahmood72 (approved):
LGTM

PR #27440, opened 2024-11-04
Increase MariaDB version and set major db upgrade flag

brown-a2:
Even though this is upgrading from 10.4 -> 10.11 it is a major update. Following this teams lead https://github.com/ministryofjustice/cloud-platform-environments/pull/26924


tmahmood72 (approved):
LGTM

PR #27443, opened 2024-11-04
Database upgrade - return option back to false

brown-a2:
Database has been updated. Now return flag back to false.


tmahmood72 (approved):
LGTM

PR #27451, opened 2024-11-04
Upgrade RDS - bump Maria DB version to latest supported

brown-a2:
Although this is semantically is a minor bump it is considered a major update and option needs to be true.


tmahmood72 (approved):
LGTM

PR #27583, opened 2024-11-06
BAVL-390 listen for the domain event video appointment cancelled in the BVLS service so the appropriate action can be taken.

matt-crowson-moj:
Adds listening for a new event to the BVLS service so the appropriate action can be taken if/when the event occurs.


tmahmood72 (approved):
LGTM

PR #27585, opened 2024-11-06
Allow access to SRE only for prod

joelstobart-moj:


FolarinOyenuga:
@joelstobart-moj is this pr still needed?

PR #27699, opened 2024-11-11
Migrate CCLF Staging database from LZ to CP

mpw5:
* Updates `snapshot_identifier` in `rds.tf` to use final snapshot from the CCLF staging database
* Renames RDS module from `rds-instance-trial` to `rds-instance-migrated` to force recreation of the database, and updates references in `rds.tf` and `irsa.tf`
* deletes unneeded commented config from `irsa.tf`


tmahmood72 (approved):
LGTM

PR #27708, opened 2024-11-11
Update RDS IDs in CCLF Staging  Prometheus rules

mpw5:
Following the migration of the database from Landing Zone, the id of the RDS instance we want to monitor has changed. This brings the alerts in line with that change.


tmahmood72 (approved):
LGTM

PR #27711, opened 2024-11-11
Add additional security group for modernisation platform ingress

MatMoore:
We want to expose the Datahub catalogue's DB to Quicksight, running in AP on the modernisation platform.


tmahmood72 (approved):
LGTM

PR #27721, opened 2024-11-11
Update 01-rbac.yaml

glenn-christmas:
Please see https://github.com/ministryofjustice/data-platform-support/issues/926


glenn-christmas:
@tom-webber Thanks for your speedy response. You've probably already worked this out but I think the problem is occurring because there is some sort of actions workflow validation step looking for a namespace that doesn't exist.

tom-webber (approved):

tom-j-smith (approved):

PR #27765, opened 2024-11-13
EMP-706: Removed staging namespace as no longer required

muyinatech:
**Changes made:**
Deleted `laa-crime-equinity-historical-data-frontend-staging` namespace as the staging environment is no longer required for the project.


tmahmood72 (approved):
LGTM

PR #27852, opened 2024-11-15
Visit scheduler RDS Pre: remove allocated storage to return namespace…

KristianGregor:
… to good state


KristianGregor:
discard

PR #27860, opened 2024-11-15
Ignore - Raising to review plan

adaviesMOJ:
Trying to confirm if changing k8s secret to new rds module would work


adaviesMOJ:
Ignore - closing as plan proved it would work for future PR.

PR #27911, opened 2024-11-18
chore: add irsa policy arn as we need it for audit connection in csip ns

ReedSoftware:


simon-mitchell (commented):
Approved from HAAR team

tmahmood72 (approved):
LGTM

PR #27919, opened 2024-11-18
Add CAS2 Bail to CAS env

adamhughes86:
We would like a dev environment for CAS2 Bail, under the existing CAS namespace because we will be calling the shared API in the future.


Rosie-Brigham (approved):

Rosie-Brigham (approved):

tmahmood72 (approved):
LGTM

PR #28063, opened 2024-11-21
EMP-723: Remove S3 bucket from UAT

tanovellino:
Removing S3 bucket (and its user policies) from UAT namespace since now this bucket is redundant


tmahmood72 (approved):
LGTM

PR #28070, opened 2024-11-21
Add CCR production pingdom tag

mpw5:
Adds the `laa_production_environment_dashboard` tag to CCR production pingdom check to add the check to the LAA dashboard.


tmahmood72 (approved):
LGTM

PR #28177, opened 2024-11-28
SDIT-2324: 🔧 Try and get logging working

petergphillips:


namespaces/live.cloud-platform.service.justice.gov.uk/hmpps-prisoner-location-dev/resources/s3.tf
```diff
@@ -106,14 +100,3 @@ resource "kubernetes_secret" "hmpps-prisoner-location_s3_bucket" {
     dso_s3_secret_access_key = aws_iam_access_key.user.secret
   }
 }
-
-resource "kubernetes_secret" "s3_logging_bucket" {
```
petergphillips, line 110: don't need this as can access the logging bucket from the primary bucket anyway

FolarinOyenuga (approved):

PR #28252, opened 2024-12-03
IR-637 & IR-639: Changes to incident reporting dev namespace. 

jonesj2193:
Setup read replica as per IR-639, added parameters for connectivity as per IR-637

### DIFF
--- a/namespaces/live.cloud-platform.service.justice.gov.uk/hmpps-incident-reporting-dev/resources/rds.tf
+++ b/namespaces/live.cloud-platform.service.justice.gov.uk/hmpps-incident-reporting-dev/resources/rds.tf
@@ -102,14 +102,33 @@ module "read_replica" {
 
   # If db_parameter is specified in source rds instance, use the same values.
   # If not specified you dont need to add any. It will use the default values.
-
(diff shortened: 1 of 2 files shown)


namespaces/live.cloud-platform.service.justice.gov.uk/hmpps-incident-reporting-dev/resources/rds.tf
```diff
@@ -21,6 +26,90 @@ module "dps_rds" {
...
+  # db_parameter = [
+  #   {
+  #     name         = "rds.force_ssl"
+  #     value        = "0"
+  #     apply_method = "immediate"
+  #   }
+  # ]
 }
 
 resource "kubernetes_secret" "dps_rds" {
```
Mjwillis, line 115: @jonesj2193 You'll need the secrets adding so we can connect to it as well

jonesj2193: Think I have done that now?

tmahmood72 (approved):
LGTM

PR #28269, opened 2024-12-04
BAVL-439 remove the migration event in prod from the new BVLS service now it is live and all data from the old service has been migrated.

matt-crowson-moj:
PROD:

Now that the new BVLS service is live and all data from the old service is migrated, we no longer need to subscribe to the migrate data event.


tmahmood72 (approved):
LGTM

PR #28360, opened 2024-12-11
added millside group to hmpps-prison-api-dev

wcdkj:
As part of https://dsdmoj.atlassian.net/browse/HMAI-83 we want to be able to query the dev instance of prison-api


petergphillips (changes requested):
please could you explain why you need access to our namespace

PR #28408, opened 2024-12-13
Reduces noise in slack alerts channel, but preserves genuine problems

andrewl:


namespaces/live.cloud-platform.service.justice.gov.uk/make-recall-decision-prod/06-prometheus.yaml
```diff
@@ -11,10 +11,10 @@ spec:
   - name: application-rules
     rules:
     - alert: nginx-5xx-error
-      expr: sum(rate(nginx_ingress_controller_requests{exported_namespace="make-recall-decision-prod", status=~"5.."}[5m]))*270 > 0
+      expr: sum(rate(nginx_ingress_controller_requests{exported_namespace="make-recall-decision-prod", status=~"5.."}[5m]))>3
```
ppmg, line 14: I've never used Prometheus, so please slap me down if needed here, but should we be keeping the `rate`part here? From [this description](https://prometheus.io/docs/prometheus/latest/querying/functions/#rate) I'm getting that it gives the rate per second averaged out over the 5m period we're giving it, but we want it to trigger if we get more than 3 over said period, no?

PR #28431, opened 2024-12-16
BAVL-476 remove SNS subcription to domain events for legacy BVLS. Now handled by re-platformed service.

matt-crowson-moj:
This change removes domain event subscription for legacy BVLS code.

These domain events are now handled by the live re-platformed version of the service.


timharrison-moj (approved):
I think there's more to delete than this, isn't there? subscription to the topic with the message types? And the secret that is created? Mind you, they may be in the domain-events namespace..

tmahmood72 (approved):
LGTM

PR #28568, opened 2025-01-02
Set backup window time

lucas-shaw:
The automated database backups were thought to run every day, however it seems they are every 2 days. Cloud platform have suggested setting a backup window to see whether this resolves the issue.


tmahmood72 (approved):
LGTM

PR #28588, opened 2025-01-02
BAVL-476 remove SNS subcription to domain events for legacy BVLS in whereabouts. Now handled by re-platformed service.

matt-crowson-moj:
This is linked to the following change for whereabouts API also [here](https://github.com/ministryofjustice/cloud-platform-environments/pull/28431).


tmahmood72 (approved):
LGTM

PR #28591, opened 2025-01-02
CLA Backend UAT add inbound ingress connection

BenMillar-MOJ:
- Allows inbound connections from `laa-access-civil-legal-aid-dev`


tmahmood72 (approved):
LGTM

PR #28598, opened 2025-01-02
BAVL-476 remove SNS subcription to offender events for legacy BVLS in whereabouts. Now handled by re-platformed service.

matt-crowson-moj:
This change removes offender event APPOINTMENT_CHANGED subscription for legacy BVLS code in whereabouts-api prod.

This event is now handled by the live re-platformed version of the service.


petergphillips (approved):
LGTM

jackstockley89 (approved):

PR #28632, opened 2025-01-06
CPR-518 remove unused S3 bucket and associated IRSA permission

markreesmoj:


markreesmoj:
Will terraform understand the dependency here and remove the permission from IRSA first? I am keen to understand this as this only exists in production and I don't want to disrupt the service account there.

tmahmood72 (approved):
LGTM

PR #28694, opened 2025-01-08
Deleting namespace for hmpps-pecs-profile-ui

jnioche-jd:
See https://dsdmoj.atlassian.net/browse/MAP-2023

TLDR; this project was never used and never will be


jnioche-jd:
Will open 3 separate PRs instead

PR #28929, opened 2025-01-17
INC-1385 [PROD] Upgrade postgres 15->16

Mjwillis:


namespaces/live.cloud-platform.service.justice.gov.uk/hmpps-incentives-prod/resources/rds.tf
```diff
@@ -55,10 +55,12 @@ module "dps_rds_replica" {
   # If any other inputs of the RDS is passed in the source db which are different from defaults,
   # add them to the replica
 
+  prepare_for_major_upgrade   = true
+
   # PostgreSQL specifics
   db_engine         = "postgres"
-  db_engine_version = "15"
-  rds_family        = "postgres15"
+  db_engine_version = "16"
```
Mjwillis, line 62: Can we upgrade replica at the same time?

jackstockley89 (approved):

namespaces/live.cloud-platform.service.justice.gov.uk/hmpps-incentives-prod/resources/rds.tf
jackstockley89: that should be fine

PR #28976, opened 2025-01-20
lpf-572: increase memory for pay for legal aid prod pod

SineadCummings:
Currently memory size is too small for larger reports


tmahmood72 (approved):
LGTM

PR #28994, opened 2025-01-21
WFP-2917 Add notification queue prod

GaReid:


FolarinOyenuga:
@GaReid is this pr still needed?

PR #29210, opened 2025-01-29
Add serviceaccount get permissions for cd sa

dteeling-moj:
Give permissions for the cd-serviceaccount to get service accounts in the delis-alfresco namespaces. This is to fix the deployment pipelines.


dteeling-moj:
Closed due to multiple namespaces being affected.

PR #29241, opened 2025-01-30
Add serviceaccount get permissions for cd sa (dev)

dteeling-moj:
Give permissions for the cd-serviceaccount to get service accounts in the hmpps-delis-alfresco-dev namespace. This is to fix the deployment pipeline.


tmahmood72 (approved):
LGTM

PR #29259, opened 2025-01-30
remove service account name from github action runner to try to fix actions issue

rgforsyth:
Removed the service account name from hmpps-find-and-refer-an-intervention-dev namespace github actions runner in order to try and fix our non-running github actions e2e tests.


tmahmood72 (approved):
LGTM

PR #29558, opened 2025-02-13
Change message retention config for visits events dlq

adaviesMOJ:
reduce from default. We don't need the message to exist longer than 1 business day.


timckt (approved):

dhirajshettymoj (approved):
LGTM

PR #29691, opened 2025-02-17
Feature/si 395 preprod

RAGQYFTJ:


RAGQYFTJ (commented):
SI-395 upgraded to 16.3 db postgres version. disabling prepare flag after upgrade

PR #29856, opened 2025-02-19
ESWE-1234 add DPR security gp and configure param groups on `dev`

gbmojo:


namespaces/live.cloud-platform.service.justice.gov.uk/hmpps-jobs-board-dev/resources/rds-postgresql.tf
```diff
@@ -26,6 +29,35 @@ module "rds" {
   is_production          = var.is_production
   namespace              = var.namespace
   team_name              = var.team_name
+
+  # add parameter group for DPS
```
rickchoijd, line 33: Is this DPR or?

rickchoijd (dismissed):
The max_slot_wal_keep_size is smaller than DPR guide, but still look sufficient for the tiny dev DB.
We need to follow the same size as in the guide in other env (preprod, `prod) DB later.

rickchoijd (approved):

PR #30012, opened 2025-02-26
CRS-2277 Add preprod rds secret to prod namespace for db refresh

srowlands-moj:
* Output RDS details to secret in prod namespace to allow refresh job to refresh rds data from prod to pre-prod


srowlands-moj:
@ministryofjustice/webops Any chance of a review please

PR #30282, opened 2025-03-10
MAP-2249 Set postgres to 16.4 for primary db

Mjwillis:


FolarinOyenuga (changes requested):
please set the db_engine_version to 16.4 as that's the version your database is currently running on

FolarinOyenuga (approved):

PR #30443, opened 2025-03-14
feat(*): add prod namespace for learner records service

malaw-moj:


namespaces/live.cloud-platform.service.justice.gov.uk/hmpps-learner-records-prod/00-namespace.yaml
```diff
@@ -0,0 +1,17 @@
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: hmpps-learner-records-prod
+  labels:
+    cloud-platform.justice.gov.uk/is-production: "false"
```
jaskaransarkaria, line 7: this should be "true"

jaskaransarkaria (approved):

PR #30533, opened 2025-03-18
introduced ingress rule to access frontend

arun-moj-laa:
This PR is created for introducing ingress rule to access mailhog front-end


namespaces/live.cloud-platform.service.justice.gov.uk/laa-smtp-mock-dev/04-networkpolicy.yaml
```diff
@@ -25,3 +24,21 @@ spec:
     - namespaceSelector:
         matchLabels:
           component: ingress-controllers
+
+---
+kind: Ingress
+metadata:
+  name: mailhog-ingress
```
andywelshmoj, line 31: a `nginx.ingress.kubernetes.io/whitelist-source-range` annotation should probably be added here to stop the service from being exposed to the public internet

PR #30626, opened 2025-03-21
Add care-arrangement-plan-prototype

josephsilcock:
Adding the cloud platform setup for the Care Arrangement Plan prototype - https://github.com/ministryofjustice/care-arrangement-plan-prototype.

Created using `cloud-platform environment prototype create`


josephsilcock:
@EarthlingDavey @EmilyHazlehurst @wilson1000
Could you please give this a quick review and post to #ask-cloud-platform

namespaces/live.cloud-platform.service.justice.gov.uk/care-arrangement-plan-prototype/resources/variables.tf
```diff
@@ -0,0 +1,76 @@
...
+
+variable "github_token" {
+  type        = string
+  description = "Required by the GitHub Terraform provider"
+  default     = ""
+}
+
+## Prototype kit variables
+
+variable "password" {
```
EarthlingDavey, line 73: I don't think we have a mechanism to securely get a password into terraform this way.

Is this going to be a password that is available to the application as an env var?

If so, could you add it as a GitHub secret and we can handle it as part of the application deployment in the application repo.

EarthlingDavey:
Hi @josephsilcock I have a couple of suggestions.

1. We would usually have a folder and namespace for pre-production, like <projectname>-dev or <projectname>-staging.
This allows us to test infra changes before going right into prod.
If this is going to be the prod namespace, would it be ok with you to suffix the namespace and folder name with `-prod` or `-production` to keep with convention?
2. I made a comment above related to passwords :)

namespaces/live.cloud-platform.service.justice.gov.uk/care-arrangement-plan-prototype/resources/variables.tf
josephsilcock: Yep, its a password used to access the prototype (doesn't need to be secure, its just to stop people randomly stumbling across it).

I was initially going to add it as a Github actions secret, but saw this was how it was set up in the cloud platform prototype setup.

I'll move it to the prototype repo - I agree it makes much more sense!

josephsilcock:
> Hi @josephsilcock I have a couple of suggestions.
> 
> 1. We would usually have a folder and namespace for pre-production, like -dev or -staging.
>    This allows us to test infra changes before going right into prod.
>    If this is going to be the prod namespace, would it be ok with you to suffix the namespace and folder name with `-prod` or `-production` to keep with convention?
> 2. I made a comment above related to passwords :)
> 3. Did you want to provision Elasticache here too?
> 
> Otherwise, looking good!

Thanks for looking!

1. This is just a prototype repository, using the gov uk prototyping kit. It won't ever go into production, as such. There are quite a few "-prototype" namespaces, so I thought it was in keeping with the convention. Happy to make the change if you think I should though?
2. I'll do this now
3. Nope, no need for elasticache in this one

EarthlingDavey:
Ah, every day's a school day! Let's keep prototype!

EarthlingDavey:
@josephsilcock any objection to me making the PR from your branch? Looks like that might be the quickest easiest solution for now.

I'll ask the Cloud Platform folks if it is actually the best way going forward.

josephsilcock:
@EarthlingDavey happy to do whatever is easiest :).

For what its worth, I'm only expecting to make a couple more PRs next week (assuming everything goes smoothly), and then this project will be done

EarthlingDavey:
Cool, looks like I'll have to close this one...

PR #31468, opened 2025-04-04
[ESWE-1289] add security and parameter groups for DPR ingestion - PROD

gbmojo:


rickchoijd (commented):
The change itself is good, but we may need to hold this change for go-live later

PR #31491, opened 2025-04-07
Add session secret for CAS2 Bail preprod

adamhughes86:
Deployments to preprod are not working for CAS2 Bail. The build says we're missing the session secret so after digging through older commits and threads I found a file that we needed to add due to the change in build process


adamhughes86:
Previous thread

https://mojdt.slack.com/archives/C06MWP0UKDE/p1732015283613529

tom-j-smith (approved):

PR #32214, opened 2025-04-15
BAVL-752 subscribe to the appointments changed domain event for the BVLS API to process.

matt-crowson-moj:
Change to subscribe to the appointments changed domain event in prod.

This is so the BVLS API can take the appropriate action for any video bookings affected by the prisoners appointments related to this event.

Associated API PR [here](https://github.com/ministryofjustice/hmpps-book-a-video-link-api/pull/459).


tmahmood72 (approved):
LGTM

PR #32227, opened 2025-04-15
Update CAS1 UI environment protection rules - prod

froddd:
`prod`: deploy review from pipeline, only main branch


tmahmood72 (approved):
LGTM

PR #32229, opened 2025-04-15
Update CAS1 UI env protection rules - preprod

froddd:
`preprod`: deploy review from pipeline, only main branch


tmahmood72 (approved):
LGTM

PR #32329, opened 2025-04-22
CDPS-1340: address security test issue for data retention period

malaw-moj:


malaw-moj:
we don't want snapshots of a read replica

PR #32562, opened 2025-04-30
DPR2-1777: Add MP DPS security group to calculate_release_dates_api_rds

koladeadewuyi-moj:
This PR adds the MP DPS security group to `calculate_release_dates_api_rds`so that the primary database instance is also accessible to the DPR service.


joelstobart-moj (approved):
LGTM

PR #32627, opened 2025-05-01
Add new event to domain events topic for visit-allocations-api - prod

adaviesMOJ:


adaviesMOJ:
Closing for now - seems this might not be needed on visits namespace side, but on nomis.

PR #33065, opened 2025-05-20
HMAI-556 Restrict API Gateway logging in development

emmalanigan:
Our API Gateway logs contain PII and other Personal data. This is not ideal and we hope that changing data_trace_enabled to false will resolve this.


tmahmood72 (approved):
LGTM

PR #33099, opened 2025-05-20
Update Team Metadata in Information Exchange Production Namespace

Mhowell494:
Update team ownership to `crime-apps-team` from `laa-ops` so that we can migrate the service over from the LAA LZ.


tmahmood72 (approved):
LGTM

PR #33239, opened 2025-05-22
Delete namespaces/live.cloud-platform.service.justice.gov.uk/lcjbexpl…

treumannabe:
…orer-dev directory

Deleting this app from hosting due to security incident while we try and work out what's in it. All contributors to the app were consultants from Faculty, and we're not sure how best to approach fixing this. In the interim, shutting down the app seems like the best start.

May revert this following discussions with project leads, but only once vulnerabilities are fixed.


treumannabe:
Sorry - I'm not familiar with any of the terminology there, can anyone advise on next steps?

PR #33446, opened 2025-05-27
Remove ECR from laa-cwa-bulk-upload-prod

muyinatech:
Changes made:

Removed ECR from laa-cwa-bulk-upload-prod, as not required.


tmahmood72 (approved):
LGTM

PR #33473, opened 2025-05-28
CAS-1681: Removed dupe RBAC for SRE and added Live CAS group

williamfalconeruk:
- Added `hmpps-community-accommodation-live` to preproduction namespace
- cleaned up duplicate SRE entry.


tmahmood72 (approved):
LGTM

PR #33474, opened 2025-05-28
MO-1698 Add github service account to MPC staging

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1698

Also, removing no longer needed service pod.


tmahmood72 (approved):
LGTM

PR #33543, opened 2025-05-29
allow resource itself for ListObject

robertsweetman:
hmpps-prisoner-location-preprod

- without the bucket's actual arn in the allowed resources list s3:ListBucket won't work
- s3:PutObject works because it's acting on arn/* i.e. the contents of the bucket

Required to check whether the new file has already been uploaded to make the workflow running the file upload idempotent
It's going to be called at 02:20 a.m. and again at 03:20 a.m. to deal with GMT/BST

Offloc file upload to the source bucket elsewhere runs at 23:20 and 00:20 UTC for the same reason


petergphillips (approved):
LGTM

PR #33648, opened 2025-06-02
Add letsencrypt cert for staging environment

mark-butler-solirius:


tmahmood72 (commented):
Please can you check the Postgress version is correct in the file rds-postgreql.tf
https://docs.aws.amazon.com/AmazonRDS/latest/PostgreSQLReleaseNotes/postgresql-versions.html#postgresql-versions-version14

PR #33664, opened 2025-06-03
update secret names prior to multiple envs

mikeAdamsMOJ:
update secret names so we can differentiate when a second env is added.


mikeAdamsMOJ:
handle a different way

PR #33736, opened 2025-06-04
`intranet-production` Add DNS for old JAC intranet domain

EarthlingDavey:
Hi @wilson1000 , can I get some input on this?

Currently there is a service that is redirecting visitors from the old JAC intranet

The old domain is jac.intranet.service.justice.gov.uk 
And the repo is at https://github.com/ministryofjustice/wp-intranet-jac

With this PR and an update to DNS records, we could bring the redirect into `intranet-production`'s ingress with a rule like 

```
# If the host is jac.intranet.service.justice.gov.uk then
# redirect to the https://intranet.justice.gov.uk/?agency=jac
if ($host = "jac.intranet.service.justice.gov.uk") {
    return 301 https://intranet.justice.gov.uk/?agency=jac;
}
```

Then we could safely archive the JAC repo, and remove it from wherever it is hosted. Ultimately reducing our surface area for security attacks.


wilson1000:
Cool, man, we change the file name to `route53-jac-redirection` instead? Make it more specific.

EarthlingDavey:
Sounds good, calling it `route53-alt` was abstracting way too early.

timckt (approved):

PR #33944, opened 2025-06-11
remove deletion protection

clmwallace:
for ccpt dev webapp


clmwallace:
I used to be involved with this app and have been asked to decommission it. I'm not an admin or member of rbac teams since I'm in a different role now. I have approval via email from the team lead in this area (Miranda.Crusco1@justice.gov.uk), saying both 'ccpt' (dev and prod) and 'hmcs-ccpt' (dev and prod) webapps should be deleted, but Miranda isn't on github to be able to approve this herself.

PR #34106, opened 2025-06-16
LSM-19 - Adjudications Live Support Namespace

theDustRoom:
Rename test namespace to live-support


namespaces/live.cloud-platform.service.justice.gov.uk/hmpps-manage-adjudications-api-live-support/00-namespace.yaml
```diff
@@ -1,10 +1,10 @@
 apiVersion: v1
 kind: Namespace
 metadata:
-  name: hmpps-manage-adjudications-api-test
+  name: hmpps-manage-adjudications-api-live-support
   labels:
     cloud-platform.justice.gov.uk/is-production: "false"
-    cloud-platform.justice.gov.uk/environment-name: "test"
+    cloud-platform.justice.gov.uk/environment-name: "live-support"
```
theDustRoom, line 7: Is this ok or are there preset environment names that we should use?

PR #34200, opened 2025-06-18
remove deletion protection crown flows (prod)

clmwallace:


clmwallace:
Apparently I'm not a member of the relevant rbac teams, but this app is owned by my team and is no longer required.

PR #34218, opened 2025-06-18
Delete namespaces/live.cloud-platform.service.justice.gov.uk/data-pla…

clmwallace:
…tform-app-judicial-wm2-dev directory


clmwallace:
I'm not an admin for this app, but it does belong to my team and is no longer required.

PR #34756, opened 2025-07-14
[CTSKF 1311] Add rule to block all egress on CCR production

jrmhaig:


jrmhaig:
Closing as this does not work as expected.

PR #34852, opened 2025-07-17
Allow MCC to communicate with CCA's development namespace

BenMillar-MOJ:
Allow Mange your Civil Cases to communicate with Civil Case API's development namespace


tmahmood72 (approved):
LGTM

PR #35004, opened 2025-07-23
CVSL-3136: Set event filter to receive no events in CVL API prod.

justin-gardiner:


justin-gardiner:
Empty arrays are not allowed in filter policy.

PR #35112, opened 2025-07-25
CPR-802 allow the pre-production service account to access the produc…

markreesmoj:
…tion S3 CHER large message bucket

This is because we have subscribed a preproduction queue to a production topic, and the messages contain a reference to the S3 bucket where large objects are stored. Consequently the preproduction service account needs get-object permission to the production S3 bucket


tmahmood72 (approved):
LGTM

PR #35192, opened 2025-07-30
Define ingress for HMPPS esupervision preprod environment

lkitching:
Create ingress which allows access from the internet to the HMPPS esupervision preprod UI.


tmahmood72 (approved):
LGTM

PR #35423, opened 2025-08-11
Added secrets manager and service account to laa-data-claims-event-service-uat

JamieBriggsDev:
Adds a secret manager and service account to laa-data-claims-event-service-uat namespace


tmahmood72 (approved):
LGTM

PR #35426, opened 2025-08-11
Lower TTL for HMICFRS domain as per requested

brown-a2:
- Stakeholder requested lowering TTL for technical support of dns changes.


tmahmood72 (approved):
LGTM

PR #35761, opened 2025-08-26
feat: remove skip and add replica

mikebell:
## Purpose

- Removes skip file
- Adds postgres replica


tmahmood72 (commented):
Don't forget to include the resource "kubernetes_secret" "read_replica" ... for the read replica

PR #35843, opened 2025-08-28
feat: remove read replica for snapshot restore

mikebell:
## Purpose

- Removes read replica for snapshot restore


tmahmood72 (approved):
LGTM

PR #36124, opened 2025-09-10
HAAR-4131: hmpps-audit-preprod - copy access control from hmpps-audit-dev for prisoner audit queue

simon-mitchell:


namespaces/live.cloud-platform.service.justice.gov.uk/hmpps-audit-preprod/resources/variables.tf
```diff
@@ -47,6 +47,11 @@ variable "github_token" {
   default     = ""
 }
 
+variable "approved_prisoner_audit_clients" {
+  type    = list(string)
+  default = ["hmpps-launchpad-dev"]
```
mtac50, line 52: @simon-mitchell this will be the issue - we can't add the approved client here until it has been added to the secret.

PR #36292, opened 2025-09-16
AP-6241: Remove pingdom alert integration id [Assure production]

jsugarman:
AP-6241: Remove pingdom alert integration id [Assure production]

Remove public #laa-civil-apply slack channel integration id
to prevent "Out of hours" 503 statuses from overwhelming
the channel. We still alert to #apply-alerts-prod.


jsugarman:
decided not to as a team

PR #36710, opened 2025-09-30
Allow civil apply uat / staging and laa ccms dev to speak to user management api dev env internally

rajindersinghsandhu:


jsugarman (approved):
LGTM

PR #36733, opened 2025-09-30
Allow prometheus metric scraping

rajindersinghsandhu:


rajindersinghsandhu:
Decided not to do this :)

PR #36956, opened 2025-10-07
CDPS-1657: Replace service accounts to swap CircleCI deployment for Github Actions in `hmpps-prisoner-profile-dev`

ushkarev:


namespaces/live.cloud-platform.service.justice.gov.uk/hmpps-prisoner-profile-dev/resources/variables.tf
```diff
@@ -30,7 +27,7 @@ variable "team_name" {
 
 variable "environment" {
   description = "The type of environment you're deploying to."
-  default     = "development"
+  default     = "dev"
```
ushkarev, line 30: this has a side effect of changing some resources’ tags, but is not destructive

PR #42166, opened 2026-05-01
fix: reduce db-restore backoffLimit from 3 to 1

raby:
Each retry restarts the pod from scratch, which re-submits a new rds_restore_database task. RDS rejects concurrent restores on the same DB, so retries just waste time waiting for the conflict error.

A single attempt is sufficient — if it fails, the activeDeadlineSeconds backstop (90 min) kills the job cleanly, and the next scheduled run at 05:45 UTC tomorrow will retry with a fresh state.


sj-williams (approved):

sablumiah:
Your PR is applying in the build: https://concourse.cloud-platform.service.justice.gov.uk/teams/main/pipelines/environments-live/jobs/apply-namespace-changes-live/builds/21067

PR #45820, opened 2026-09-16
updated mdt namespace to use the hmpps prefix for dev env

kurt-uk:
updated namespace to use hmpps prefix


kurt-uk:
closing as not needed

